PR #1, opened 2020-05-27
Fix clippy warnings & rustfmt

Scotsguy:
In two commits for cherrypicking purposes.
Test the camelcasing before merging


Geometrically:
Panic on run:
`thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("missing field `is_default`", line: 1, column: 768)', src/routes/search.rs:206:47
`

PR #6, opened 2020-06-23
Implemented a dark theme

emilyploszaj:
* Migrated most hard coded colors that matter to css variables
* Created a second set for a dark theme
* Set up the ability to theme switch, inferring from system theme


templates/search.hbs
```diff
@@ -5,6 +5,8 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
 
+	<script src="/static/js/main.js"></script>
```
Geometrically, line 8: Could you move these to the bottom of the body, where the rest of the scripts are?

templates/mod-page.hbs
```diff
@@ -5,6 +5,8 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
 
+	<script src="/static/js/main.js"></script>
```
Geometrically, line 8: See other comment

templates/mod-create.hbs
```diff
@@ -5,6 +5,8 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
 
+	<script src="/static/js/main.js"></script>
```
Geometrically, line 8: See other comment

emilyploszaj:
It's better for the script to be loaded before the css is so it think the theme has not changed and doesn't occasionally transition colors on page load

Geometrically:
Makes sense.

PR #7, opened 2020-06-25
Improve styling for narrower screens

Aeledfyr:
This is mainly some changes to make the styles of search results work better with narrower screens.

* The category badges on each result collapse to just an icon (with title text) when the screen is too narrow.
* Adds a text label for the Forge/Fabric icons.
* Adds a message for when a query returns no results.


Aeledfyr:
The next step for mobile support would be to hide the sidebars in some way when there isn't enough space.  I'm not sure how to go about that, but the category and version listings would probably have to go behind a menu on mobile.

I'm not sure about the versions list, but the category list and loader list could be drop down like menus.

templates/search-results.hbs
```diff
@@ -3,113 +3,135 @@
...
+                            </svg><!--
+                            --><span>Forge</span>
+                        </div>
                     {{/contains}}
                     {{#contains this.keywords "fabric"}}
-                        <img alt="fabric" src="/static/images/icon/fabric.png"/>
+                        <div class="loader-fabric" title="Fabric">
+                            <img alt="fabric" src="/static/images/icon/fabric.png"/><!--
+                            --><span>Fabric</span>
+                        </div>
```
Geometrically, line 48: The text should be removed IMO. It doesn't add much, as you could see the logo icons already and it looks cluttered

templates/search-results.hbs
```diff
@@ -3,113 +3,135 @@
...
         </div>
     </div>
+{{else}}
+    <div class="search-error">
+        {{#if ../query.q}}
+        <h2>No results found for query <code>"{{../query.q}}"</code></h2>
+        {{else}}
+        <h2>No results found</h2>
+        {{/if}}
+    </div>
```
Geometrically, line 136: This part is super great but if the search query is long enough the div will overflow. Not a big problem though, but should be an easy fix
Screenshot_20200625_175347

templates/search-results.hbs
Aeledfyr: I can remove the text, but the icons are very small, so it's hard to understand them if you don't already know the icons for forge and fabric.  Since I added a title text to the icon, it should be fine to remove the text to reduce clutter, but it may reduce readability.

Also, is the forge icon an inline svg just for changing the color with themes?  It makes it harder to align properly, and seems like using a normal image combined with css filters or clip masks would be more efficient.

Geometrically: Yeah, the inline svg is only there to change the color. If you have another way to do it, that would be great! It would save a lot of bandwidth

Aeledfyr: I've replaced the inline SVG with a SVG `<use href="#...">` element, so the icon is only inlined once.  For some reason this works with the original theming CSS, so it's only changes to the templates.

Geometrically:
Looks good. Thanks for your help and the pull request!

PR #12, opened 2020-06-27
Make category badges more generic

joaoh1:
This PR does a bunch of little things that end cleaning up the style behind the category badges a little:
- The only PNG badges (fabric.png and cursed.png) have been resized in order to be a perfect square and then optimized with oxipng.
- Now that the PNGs are perfect squares, the badges are 23px wide, not 23px tall.
- The specific rules for the Forge and the cursed category's badges have been removed.


Aeledfyr:
Using [this app](https://squoosh.app/editor), I can get it down to 1.3KiB with oxipng by reducing the color palette down to 16 colors.  Since this will be a 23x23px icon, the slight dithering that causes doesn't seem like it should be a problem.  (Without dithering, it goes down to 800 bytes). 
image

joaoh1:
> 
> 
> Using [this app](https://squoosh.app/editor), I can get it down to 1.3KiB with oxipng by reducing the color palette down to 16 colors. Since this will be a 23x23px icon, the slight dithering that causes doesn't seem like it should be a problem. (Without dithering, it goes down to 800 bytes).
> image

I don't really want to risk any sort of quality loss, I think just feeding it to oxipng without reducing the pallette or sacrificing anything is good enough.

Geometrically:
The tater is pretty simple, the image Aeled linked didn't even lose much quality.

Geometrically:
Merging, we can discuss minifying PNGs in a separate issue

Issue #16, opened 2020-06-28
Remove all frontend in the application


Redblueflame:
Following the decision to move to separate repos for frontend and backend, I'll clean everything up

Geometrically:
This will be fun

Redblueflame:
Loads of fun ! 🤣

Redblueflame:
Reopening it to return at normal state

PR #18, opened 2020-06-29
Rewrite the app

### LINKED ISSUE #2832: Remove all frontend in the application
Following the decision to move to separate repos for frontend and backend, I'll clean everything up

Redblueflame:
Rewrite everything, so this project only contains the api side of things, and the client has moved to [dungeon](https://github.com/modrinth/dungeon)

This what to do for the rewrite

- [x] Remove everything related to the UI
- [ ] Add new routes for the search calls
- [ ] Add a database agnostic abstraction for the database
- [ ] Add a scheduled task for curseforge indexing
- [x] Move everything related to search from the route to a specific module


src/search/models.rs
```diff
@@ -0,0 +1,82 @@
...
+    pub attachments: Vec<Attachment>,
+    pub website_url: String,
+    pub summary: String,
+    pub download_count: f32,
+    pub categories: Vec<Category>,
+    pub game_version_latest_files: Vec<CurseVersion>,
+    pub date_created: String,
+    pub date_modified: String,
+    pub game_slug: String,
+}
```
Geometrically, line 45: Might want to move this to the curseforge indexer, as this is curseforge code

src/search/models.rs
```diff
@@ -0,0 +1,82 @@
...
+    pub page_url: String,
+    pub icon_url: String,
+    pub author_url: String,
+    pub date_created: String,
+    pub created: i64,
+    pub date_modified: String,
+    pub updated: i64,
+    pub latest_version: String,
+    pub empty: String,
+}
```
Geometrically, line 74: I think this should go in the models module or just in the search class. It seems sort of pointless for a models file in the search if we have a models module.

Geometrically (changes requested):

docker-compose.yml
```diff
@@ -0,0 +1,33 @@
+version: '3'
+services:
+  mongo:
+    image: mongo
+    restart: always
+    ports: 
+      - 27017:27017
+    environment:
+      MONGO_INITDB_ROOT_USERNAME: toor
+      MONGO_INITDB_ROOT_PASSWORD: fabricateadmin
```
Geometrically, line 10: Should probably be changed to labrinth or modbrinth admin, since we aren't called fabricate anymore

src/search/models.rs
Redblueflame: I'm moving it, it's even used in the returned value I think

src/search/indexing/local_import.rs
```diff
@@ -0,0 +1,54 @@
...
+use futures::StreamExt;
+use crate::database::models::Item;
+use crate::database::{Version, Mod};
+
+pub async fn index_local(client: mongodb::Client) -> Result<Vec<SearchMod>,  IndexingError> {
+    info!("Indexing local mods!");
+
+    let mut docs_to_add: Vec<SearchMod> = vec![];
+
+    let db = client.database("fabricate");
```
Geometrically, line 14: Database should be renamed to match name change

Geometrically (changes requested):

src/search/models.rs
Redblueflame: Yup, everything was just separated to be more clear

docker-compose.yml
Redblueflame: It's really here to be a testing system

src/search/indexing/local_import.rs
Redblueflame: I'm modifying it

src/search/models.rs
Geometrically: Alright

src/search/indexing/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+    FloatParsingError(#[from] std::num::ParseFloatError),
+    #[error("Error while parsing float")]
+    IntParsingError(#[from] std::num::ParseIntError),
+    #[error("Error while parsing GSON")]
+    DatabaseError(#[from] DatabaseError)
+}
+
+pub async fn index_mods(db: mongodb::Client) -> Result<(), IndexingError>{
+    // Check if the index exists
+    let client = Client::new("http://localhost:7700", "");
```
Aeledfyr, line 34: This should probably use `dotenv` with a default instead of hardcoding the url

src/search/mod.rs
```diff
@@ -0,0 +1,49 @@
+use crate::search::models::{SearchRequest, SearchMod};
+use meilisearch_sdk::client::Client;
+use meilisearch_sdk::search::Query;
+
+pub mod indexing;
+pub mod models;
+
+fn search(info: &SearchRequest) -> Vec<SearchMod> {
+    let client = Client::new("http://localhost:7700", "");
```
Aeledfyr, line 9: Same as above, should use `dotenv` for url

Aeledfyr (changes requested):

src/search/indexing/mod.rs
Geometrically: will do

src/search/mod.rs
Geometrically: will do

Redblueflame:
Please don't merge tonight, there is still issues to fix.

Geometrically:
I won't

Geometrically:
Fixes #16

Geometrically:
woops!!!

PR #28, opened 2020-06-30
Create a schema for the API

Aeledfyr:
This adds structures for an API schema.  It currently covers mods, their versions, and some of the ids that may be used.

The `Id` structs will need custom `serde` implementations for parsing base62 (unless there's a library that deals with that).  The `Team` structure is still unfinished, and more specifics will need to be worked out for most of the structures.

### DIFF
--- a/src/models/ids.rs
+++ b/src/models/ids.rs
@@ -1,7 +1,53 @@
 use thiserror::Error;
 
 pub use super::mods::{ModId, VersionId};
-pub use super::teams::{UserId, TeamId};
+pub use super::teams::{TeamId, UserId};
+
+/// Generates a random 64 bit integer that is exactly `n` characters
+/// long when encoded as base62.
+///
+/// Uses `rand`'s thread rng on every call.
+///
+/// # Panics
+///
+/// This method panics if `n` is 0 or greater than 11, since a `u64`
+/// can only represent up to 11 character base62 strings
+pub fn random_base62(n: usize) -> u64 {
+    use rand::Rng;
+    assert!(n > 0 && n <= 11);
+    let mut rng = rand::thread_rng();
+    // gen_range is [low, high): max value is `MULTIPLES[n] - 1`,
+    // which is n characters long when encoded
+    rng.gen_range(MULTIPLES[n - 1], MULTIPLES[n])
+}
+
+/// Generates a random 64 bit integer that is exactly `n` characters
+/// long when encoded as base62, using the given rng.
+///
+/// # Panics
+///
+/// This method panics if `n` is 0 or greater than 11, since a `u64`
+/// can only represent up to 11 character base62 strings
+pub fn random_base62_rng<R: rand::RngCore>(rng: &mut R, n: usize) -> u64 {
+    use rand::Rng;
+    assert!(n > 0 && n <= 11);
+    rng.gen_range(MULTIPLES[n - 1], MULTIPLES[n])
+}
+
+const MULTIPLES: [u64; 12] = [
+    1,
+    62,
+    62 * 62,
+    62 * 62 * 62,
+    62 * 62 * 62 * 62,
+    62 * 62 * 62 * 62 * 62,
+    62 * 62 * 62 * 62 * 62 * 62,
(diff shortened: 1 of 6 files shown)


src/models/mods.rs
```diff
@@ -1,5 +1,71 @@
...
+#[derive(Serialize, Deserialize)]
+#[serde(transparent)]
+pub struct VersionId(String);
+
+// TODO: permissions, role names, etc
+#[derive(Serialize, Deserialize)]
+pub struct Team {
+    users: Vec<(String, UserId)>,
+}
+
```
cfanoulis, line 24: Move this struct to a different file, and simply reference a team ID here?

src/models/mods.rs
```diff
@@ -1,5 +1,71 @@
...
+    pub title: String,
+    pub description: String,
+    pub published: DateTime<Utc>,
+
+    pub downloads: u32,
+    pub categories: Vec<String>,
+    pub versions: Vec<VersionId>,
+
+    pub body_url: String,
```
cfanoulis, line 38: what does this do

Aeledfyr: I don't really know, it's from the database model.  I would assume that it's a link to the main page of the mod, which doesn't make sense anymore with the API being separated from the front end.

src/models/mods.rs
Aeledfyr: Yeah, and I should also move the Id structs to their own file as well.

src/models/mods.rs
cfanoulis: I think its safe to not include this on the schema then

src/models/ids.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    const BASE62_CHARS: [u8; 62] = [
+        b'0', b'1', b'2', b'3', b'4', b'5', b'6', b'7', b'8', b'9', b'A', b'B', b'C', b'D', b'E',
+        b'F', b'G', b'H', b'I', b'J', b'K', b'L', b'M', b'N', b'O', b'P', b'Q', b'R', b'S', b'T',
+        b'U', b'V', b'W', b'X', b'Y', b'Z', b'a', b'b', b'c', b'd', b'e', b'f', b'g', b'h', b'i',
+        b'j', b'k', b'l', b'm', b'n', b'o', b'p', b'q', b'r', b's', b't', b'u', b'v', b'w', b'x',
+        b'y', b'z',
+    ];
+
+    fn to_base62(mut num: u64) -> String {
```
Redblueflame, line 107: I think it's better to use a `Vec<u8>` for that, if we have less data to store, we save some space, and it's more extensible than a hard limit at 64 bits

src/models/mod.rs
```diff
@@ -1,2 +1,4 @@
 pub mod error;
+pub mod ids;
```
Redblueflame, line 2: Creating an ids file isn't a great idea. From my opinion, using the specific file, so `teams` for the team for example is more readable than just a file where you put every single id in ta single file

src/models/ids.rs
Aeledfyr: It would be more extensible, but it would take up more space, at least on the rust side, and would make the implementation slower and more complicated.  We would effectively have to use a BigInteger as an input, which would be significantly slower.  A 64 bit integer can already handle up to around 11 character base62 strings, which should be good enough for most of our ids — Youtube uses 11 characters of base64.

src/models/mod.rs
Aeledfyr: I can move the individual id definitions into the relevant submodules, but then most files will have to import the ids from multiple different modules.  I would also have to keep the `ids` module for the `Base64Id` implementation.  If you're fine with that, I'll do it, but it will increase the amount of imports in every file that uses them.

Redblueflame: Increasing the amount of imports isn't a problem. A cleaner way is to move the base62 encoding to the mod.rs file sinc it's a utility for all submodules.

src/models/ids.rs
Redblueflame: Yup !

Redblueflame:
Merging

PR #30, opened 2020-07-01
feat: Add label syncing.

cfanoulis:
This will allow us to have similar labels across all of our projects, and makes editing far easier.

PR incoming in knossos as well

(and yes, there's a discord.xml, git whined about it so here you have it)


Geometrically:
What exactly do the namespaces do?

cfanoulis:
my bad for that, fat finger'd

Nanespace refers to the area of code the issue/pr touches. f.e. the mod endpoints fall under the Mod namespace

PR #32, opened 2020-07-01
Backblaze Driver

Geometrically:
This PR adds a simple backblaze driver for support for uploading and deleting files to the Backblaze B2 Bucket.

It has tests for checking if the code is working correctly too.


src/file_hosting/authorization.rs
```diff
@@ -0,0 +1,66 @@
...
+    let req = reqwest::Client::new()
+        .get("https://api.backblazeb2.com/b2api/v2/b2_authorize_account")
+        .header(reqwest::header::CONTENT_TYPE, "application/json")
+        .header(reqwest::header::AUTHORIZATION, formatted_key)
+        .send()
+        .await?;
+
+    let text = &req.text().await?;
+
+    Ok(serde_json::from_str(text)?)
```
Aeledfyr, line 49: Reqwest has serde integration, so you can use `Ok(req.json()?)` instead if you add the `"json"` feature.  This is the same will all other instances of this, but I won't comment on them all.

Redblueflame: +1

src/file_hosting/delete.rs
```diff
@@ -0,0 +1,27 @@
+use serde::{Deserialize, Serialize};
+use serde_json::json;
+use crate::file_hosting::{AuthorizationData, FileHostingError};
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct DeleteData {
```
Redblueflame, line 7: I think something like `DeleteFileData`, this is too generic

src/file_hosting/mod.rs
```diff
@@ -0,0 +1,46 @@
...
+
+#[derive(Error, Debug)]
+pub enum FileHostingError {
+    #[error("Error while accessing the data from remote")]
+    RemoteWebsiteError(#[from] reqwest::Error),
+    #[error("Error while serializing or deserializing JSON")]
+    SerDeError(#[from] serde_json::Error),
+}
+
+#[cfg(test)]
```
Redblueflame, line 27: The tests are including too much different functions, separing them into multiple ones and one final that test their integration together is needed

src/file_hosting/upload.rs
```diff
@@ -0,0 +1,35 @@
+use serde::{Deserialize, Serialize};
+use crate::file_hosting::FileHostingError;
+use crate::file_hosting::authorization::UploadUrlData;
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct UploadData {
```
Redblueflame, line 7: Same as before, `UploadFileData`

src/file_hosting/mod.rs
Geometrically: the problem with that is all of the methods depend on each other so doing that is basically impossible

src/file_hosting/authorization.rs
Geometrically: This method doesn't exist in the async impl of reqwest.

Aeledfyr: Is this not the correct method?  You have to enable the "json" feature, but it's there.  https://docs.rs/reqwest/0.10.6/reqwest/struct.Response.html#method.json

Geometrically: Fixed.

PR #33, opened 2020-07-02
Improve error handling

Aeledfyr:
This sets up error handling for actix routes, and removes unneeded unwraps.  New error types should implement actix's `actix_web::ResponseError`, and for some reason returning a `Result` from an actix route only works when the function is `async`.


src/search/mod.rs
```diff
@@ -10,24 +13,51 @@ pub mod indexing;
...
 pub enum SearchError {
-    #[error("Error while connection to the MeiliSearch database")]
-    IndexDBError(),
-    #[error("Error while connecting to the local server")]
-    LocalDatabaseError(#[from] mongodb::error::Error),
-    #[error("Error while accessing the data from remote")]
-    RemoteWebsiteError(#[from] reqwest::Error),
-    #[error("Error while serializing or deserializing JSON")]
+    #[error("Error while connecting to the MeiliSearch database")]
+    IndexDBError(meilisearch_sdk::errors::Error),
```
Redblueflame, line 17: Adding `#[from]` is recommended to prevent inconsistency

src/search/mod.rs
```diff
@@ -10,24 +13,51 @@ pub mod indexing;
...
+            SearchError::DatabaseError(..) => StatusCode::INTERNAL_SERVER_ERROR,
+            SearchError::IndexDBError(..) => StatusCode::INTERNAL_SERVER_ERROR,
+            SearchError::CurseforgeImportError(..) => StatusCode::INTERNAL_SERVER_ERROR,
+            SearchError::SerDeError(..) => StatusCode::BAD_REQUEST,
+            SearchError::FloatParsingError(..) => StatusCode::BAD_REQUEST,
+            SearchError::IntParsingError(..) => StatusCode::BAD_REQUEST,
+        }
+    }
+
+    fn error_response(&self) -> HttpResponse {
```
Redblueflame, line 45: The json `error` should be a snake_case message that is consistent between endpoints

src/search/mod.rs
Aeledfyr: Neither of the two that don't have `#[from]` can/should have it.  Meilisearch's `Error` struct doesn't implement `std::error::Error`, so it can't have a `#[from]` impl, and the `CurseforgeImportError` is more specific and doesn't apply to all `reqwest` errors, so it shouldn't have it.

PR #34, opened 2020-07-12
Mod Creation

Geometrically:
This is a PR adding a route for mod creation. As it is a draft, it needs a lot of work. Please leave review comments if you would like anything changed.

Todo list before merge:

- [x] Add support for multiple files to be attached to one version
- [x] Create environmental variable to set CDN url
- [x] Use database models instead of response ones
- [x] Add File extension checking to prevent abuse
- [x] Add support for uploading icons
- [x] Remove unwraps and replace with proper error handling


src/models/mods.rs
```diff
@@ -27,8 +27,12 @@ pub struct Mod {
 
     /// The title or name of the mod.
     pub title: String,
+    ///The URL of the icon of the mod
+    pub icon_url: String,
     /// A short description of the mod.
     pub description: String,
+    /// A long description of the mod.
```
Aeledfyr, line 34: Should be "A link to a long description of the mod."

src/search/indexing/mod.rs
```diff
@@ -22,7 +22,7 @@ pub async fn index_mods(db: mongodb::Client) -> Result<(), SearchError> {
         .parse()
         .unwrap()
     {
-        docs_to_add.append(&mut index_curseforge(1, 400000).await?);
+        docs_to_add.append(&mut index_curseforge(1, 200000).await?);
```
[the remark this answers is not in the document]
Geometrically: Oops. I'll open up an issue for this. if it's over 200,000 the payload will be too large.

src/models/mods.rs
```diff
@@ -73,13 +73,13 @@ pub struct Version {
     pub files: Vec<VersionFile>,
     /// A list of mods that this version depends on.
     pub dependencies: Vec<ModId>,
```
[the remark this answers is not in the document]
Geometrically: Great idea. Theoretically it could just link to VersionId since those should be unique throughout all versions

src/search/indexing/mod.rs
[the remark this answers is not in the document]
Geometrically: #35

src/routes/mod_creation.rs
```diff
@@ -0,0 +1,228 @@
...
+                        .unwrap();
+
+                        icon_url = format!("cdnurl/{}", upload_data.file_name)
+                    } else if file_extension == "jar".to_string() {
+                        let initial_version_data = create_data
+                            .initial_versions
+                            .iter()
+                            .position(|x| x.file_names.contains(&file_name.to_string()));
+
+                        if let Some(version_data_index) = initial_version_data {
```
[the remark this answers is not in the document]
Geometrically: Will do. The code is currently long and extremely nested. I think splitting up it into several functions would help

src/routes/mod_creation.rs
```diff
@@ -0,0 +1,228 @@
...
+        upload_file(
+            upload_url.get_ref().clone(),
+            "text/html".to_string(),
+            body_url.clone(),
+            parsed_body.into_bytes(),
+        )
+        .await
+        .unwrap();
+
+        //TODO checks to see if randomly generated ids match
```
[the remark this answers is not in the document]
Geometrically: It is generated at the top of the route since the version files need a mod id to be linked to

src/routes/mod_creation.rs
```diff
@@ -0,0 +1,228 @@
...
+
+                            let body_url =
+                                format!("data/{}/changelogs/{}/body.html", mod_id.0, version_id.0);
+
+                            upload_file(
+                                upload_url.get_ref().clone(),
+                                "text/html".to_string(),
+                                body_url.clone(),
+                                parsed_body.into_bytes(),
+                            )
```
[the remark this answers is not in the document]
Geometrically: What do you think of parsing the markdown through the frontend and storing the markdown in the CDN

src/models/ids.rs
```diff
@@ -84,12 +84,27 @@
...
+    ($struct:ty) => {
+        impl std::fmt::Display for $struct {
+            fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                f.write_str(&base62_impl::to_base62(self.0))
+            }
+        }
+    };
+}
+impl_base62_display!(Base62Id);
+
```
Redblueflame, line 97: Concern

src/models/mods.rs
```diff
@@ -27,8 +27,12 @@ pub struct Mod {
 
     /// The title or name of the mod.
     pub title: String,
+    ///The URL of the icon of the mod
+    pub icon_url: String,
     /// A short description of the mod.
     pub description: String,
+    /// A long description of the mod.
+    pub body_url: String,
```
Redblueflame, line 35: Why url ? Will this be on the backblaze bucket ?

src/models/mods.rs
```diff
@@ -96,15 +96,15 @@ pub struct FileHash {
     pub hash: String,
 }
 
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, Clone)]
 pub enum VersionType {
```
Redblueflame, line 100: Making this a simple string that is possible to edit should be better. It allows for more channels and avoids the conflicts of release settings

src/routes/mod_creation.rs
```diff
@@ -0,0 +1,228 @@
...
+    FileHash, GameVersion, Mod, ModId, Version, VersionFile, VersionId, VersionType,
+};
+use crate::models::teams::{Team, TeamId, TeamMember};
+use actix_multipart::{Field, Multipart};
+use actix_web::web::Data;
+use actix_web::{middleware, post, HttpResponse};
+use bson::Bson;
+use chrono::Utc;
+use futures::StreamExt;
+use mongodb::Client;
```
Redblueflame, line 13: The models module should **not** include database stuff. This module is only for API returned values, present in the body.

src/routes/mod_creation.rs
```diff
@@ -0,0 +1,228 @@
...
+    pub categories: Vec<String>,
+    /// An optional link to where to submit bugs or issues with the mod.
+    pub issues_url: Option<String>,
+    /// An optional link to the source code for the mod.
+    pub source_url: Option<String>,
+    /// An optional link to the mod's wiki page or other relevant information.
+    pub wiki_url: Option<String>,
+}
+
+#[post("api/v1/mod")]
```
Redblueflame, line 53: As said above, move the logic to the routes module

src/routes/mod_creation.rs
```diff
@@ -0,0 +1,228 @@
...
+    mut payload: Multipart,
+    client: Data<Client>,
+    upload_url: Data<UploadUrlData>,
+) -> HttpResponse {
+    let db = client.database("modrinth");
+
+    let mods = db.collection("mods");
+    let versions = db.collection("versions");
+
+    let mod_id = ModId(random_base62(8));
```
Redblueflame, line 64: Maybe make this a env or constant shared with everything ?

Redblueflame, line 64: Maybe make this a env or constant shared with everything ?

src/routes/mod_creation.rs
```diff
@@ -0,0 +1,228 @@
...
+    let version_ids: Vec<VersionId> = vec![];
+
+    let mut mod_create_data: Option<ModCreateData> = None;
+    let mut icon_url = "".to_string();
+
+    while let Some(item) = payload.next().await {
+        let mut field: Field = item.expect("Error while splitting payload");
```
Redblueflame, line 71: No expect, please use the error type for that, we don't want a worker to panic

src/models/mods.rs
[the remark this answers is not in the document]
Aeledfyr: The issue with that is that minor patches will be in new versions, so depending on a specific version of the mod prevents bug fixes from being applied to that mod.  That also cause issues because I assume most mods don't support having multiple versions of themselves loaded at once - we may end up with a dependency hell sort of situation unless we deal with this carefully.

src/routes/mod_creation.rs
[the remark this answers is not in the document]
Aeledfyr: Then the mod id generation comment should probably be at the top as well, so that it is implemented before `mod_id` is used.

src/models/ids.rs
Aeledfyr: If you want to write out the exact same code for every `Id` struct, go ahead.  I made this as a quick patch for Geometrically so that `Display` would work for the ids.

src/models/mods.rs
Aeledfyr: We could use a string, but we would still want a limited list of allowed version types - we don't want to have to deal with people labeling their versions with inappropriate language, or labeling their versions with types that make no sense to most people.

src/routes/mod_creation.rs
Aeledfyr: That method is currently in the `routes::mod_creation` module.

src/routes/mod_creation.rs
Aeledfyr: Is that about the mod id?  That is randomly generated for each new mod.

src/routes/mod_creation.rs
Aeledfyr: Error handling is one of the current TODO items, all of the `expect`s and `unwrap`s are temporary.

src/models/mods.rs
```diff
@@ -73,13 +73,13 @@ pub struct Version {
     pub files: Vec<VersionFile>,
     /// A list of mods that this version depends on.
     pub dependencies: Vec<ModId>,
-    /// A list of versions of Minecraft that this version of the mod supports.
-    pub game_versions: Vec<GameVersion>,
```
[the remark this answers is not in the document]
Geometrically: They are specified in the Version already if that's what you mean. I will add an enum for different types of sided mods

src/models/mods.rs
```diff
@@ -38,10 +42,6 @@ pub struct Mod {
     pub categories: Vec<String>,
     /// A list of ids for versions of the mod.
     pub versions: Vec<VersionId>,
-
-    /// The latest version of the mod.
-    pub latest_version: Version,
```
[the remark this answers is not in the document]
Geometrically: Does it need to be in the struct though? If an API client requests a mod they would be getting the version list anyways. I think a better approach is to make a struct with a "minified" version of the mod with less information.

src/models/mods.rs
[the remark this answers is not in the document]
Geometrically: Agreed. We could have it support a "Major" version of the mod i.e. 1.x instead of 1.1.x. However not all mod authors will follow this rule. I think it's probably best for the modpack authors to handle this and for the modder to just specify a specific version that they know will work.

src/routes/mod_creation.rs
```diff
@@ -0,0 +1,228 @@
...
+    /// A short description of the mod.
+    pub mod_description: String,
+    /// A long description of the mod, in markdown.
+    pub mod_body: String,
+    /// A list of initial versions to upload with the created mod
+    pub initial_versions: Vec<InitialVersionData>,
+    /// The team of people that has ownership of this mod.
+    pub team_members: Vec<TeamMember>,
+    /// A list of the categories that the mod is in.
+    pub categories: Vec<String>,
```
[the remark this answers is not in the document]
Geometrically: I forgot to add the mod loader stuff, I'll make it an enum too.

src/routes/mod_creation.rs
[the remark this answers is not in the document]
Geometrically: Gotcha. Do you think this should be moved inside the helper method for random ID generation? I don't see where we would generate a mod ID without checking if it exists already.

src/models/mods.rs
Geometrically: Yeah. Storing it in the DB is inefficient and bloats the size of the database by a ton.

src/routes/mod_creation.rs
Geometrically: Yep, like I said in the todo list, I'll be moving them into the database models.

src/routes/mod_creation.rs
Geometrically: I think RedBlue is talking about the tables. It's not necessary for them to be an enviorment variable since the tables are generated by the program and don't really need user input.

Geometrically: See above comment.

src/routes/mod_creation.rs
Geometrically: Will resolve once I get around to error handling.

src/models/mods.rs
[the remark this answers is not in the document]
Aeledfyr: The latest version doesn't have to be in the struct itself, it could just be a `VersionId` instead of a full `Version`.  It would add a second request for most uses of the API, though: search listings show the latest supported MC version, mod pages show the last released files, etc.  When a client requests a mod, it only gets a list of the `VersionId`s of every version (which we may want to limit to the last 10 or so versions).  Finding the latest version if it isn't specified requires making a request for every version of the mod to check the info for each one.

src/file_hosting/mod.rs
```diff
@@ -30,6 +30,7 @@ mod tests {
 
     #[actix_rt::test]
     async fn test_authorization() {
```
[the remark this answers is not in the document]
Geometrically: It won't

src/database/models/version_item.rs
```diff
@@ -3,19 +3,37 @@ use crate::database::Result;
...
-    pub files_path: Vec<String>,
+    pub name: String,
+    pub number: String,
+    pub changelog_url: Option<String>,
     pub date_published: String,
-    pub author: String,
     pub downloads: i32,
-    pub dependencies: Vec<String>,
+    pub files: Vec<VersionFile>,
+    pub dependencies: Vec<i32>,
```
Aeledfyr, line 17: What specifically does this ID reference?  It may be useful to add documentation / doc comments, since we don't have the type wrappers.

src/database/models/version_item.rs
```diff
@@ -3,19 +3,37 @@ use crate::database::Result;
...
+//TODO: Files should probably be moved to their own table
 #[derive(Deserialize, Serialize)]
 pub struct Version {
     pub version_id: i32,
     pub mod_id: i32,
-    pub title: String,
-    pub changelog_path: String,
-    pub files_path: Vec<String>,
+    pub name: String,
+    pub number: String,
```
Aeledfyr, line 12: Bike-shedding, but `version_number` may be a more descriptive name for this.

src/models/mods.rs
```diff
@@ -72,14 +73,16 @@ pub struct Version {
     /// A list of files available for download for this version.
     pub files: Vec<VersionFile>,
     /// A list of mods that this version depends on.
-    pub dependencies: Vec<ModId>,
-    /// A list of versions of Minecraft that this version of the mod supports.
-    pub game_versions: Vec<GameVersion>,
+    pub dependencies: Vec<VersionId>,
+    /// The loaders that this version works on
+    pub loaders: Vec<String>,
```
Aeledfyr, line 78: We should keep this field, but we should probably also add a list of the supported loaders to each `VersionFile`, as some mods have multiple files for each version, each supporting a certain mod loader.

Also for mod loader compatibility, it may be a good idea to specify the dependencies down to a specific file, in case the mod loaders don't match.

src/routes/mod_creation.rs
```diff
@@ -0,0 +1,401 @@
...
+            },
+            description: &self.to_string(),
+        })
+    }
+}
+
+#[derive(Serialize, Deserialize, Clone)]
+struct InitialVersionData {
+    pub file_indexes: Vec<i32>,
+    pub version_slug: String,
```
Aeledfyr, line 69: `version_slug` should match the naming of the field in the schema, so `version_number`

src/routes/mod_creation.rs
```diff
@@ -0,0 +1,228 @@
...
+use pulldown_cmark::{Options, Parser};
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Clone)]
+struct InitialVersionData {
+    pub file_names: Vec<String>,
+    pub version_slug: String,
+    pub version_title: String,
+    pub version_body: String,
+    pub dependencies: Vec<ModId>,
```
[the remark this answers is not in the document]
Aeledfyr: I believe this is still an issue - it should be a `Vec<VersionId>`, as that is what is is being interpreted as later in the code.

src/routes/mod_creation.rs
```diff
@@ -0,0 +1,401 @@
...
+
+    let db = client.database("modrinth");
+
+    let mods = db.collection("mods");
+    let versions = db.collection("versions");
+
+    let mut mod_id = ModId(random_base62(8));
+
+    //Check if ID is unique
+    loop {
```
Aeledfyr, line 119: We may want to add a second condition that it can't take more than 10 or 20 cycles to prevent a potential infinite loop.  If it fails, log an error and break out so that the server doesn't stall.

src/routes/mod_creation.rs
```diff
@@ -0,0 +1,401 @@
...
+                .map_err(CreateError::InvalidUtf8Input)?;
+
+                if let Some(create_data) = &mod_create_data {
+                    if name == "icon" {
+                        if let Some(ext) = get_image_content_type(file_extension) {
+                            let upload_data = upload_file(
+                                upload_url.get_ref().clone(),
+                                ext,
+                                format!("mods/icons/{}/{}", mod_id.0, file_name),
+                                data.to_vec(),
```
Aeledfyr, line 179: Uploading a file should borrow a slice rather than requiring an extra allocation.

src/routes/mod_creation.rs
```diff
@@ -0,0 +1,401 @@
...
+                                            algorithm: "sha1".to_string(),
+                                            hash: upload_data.content_sha1,
+                                        }],
+                                        url: format!("{}/{}", cdn_url, upload_data.file_name),
+                                    });
+                                }
+                                None => {
+                                    let mut version_id = VersionId(random_base62(8));
+                                    //Check if ID is unique
+                                    loop {
```
Aeledfyr, line 237: Same as above, infinite loop prevention - we should never be that close to full, but we want errors rather than stalling.

src/routes/mod_creation.rs
```diff
@@ -0,0 +1,401 @@
...
+                                    let upload_data = upload_file(
+                                        upload_url.get_ref().clone(),
+                                        "application/java-archive".to_string(),
+                                        format!(
+                                            "{}/{}/{}",
+                                            create_data.mod_namespace.replace(".", "/"),
+                                            version_data.version_slug,
+                                            file_name
+                                        ),
+                                        (&data).to_owned().to_vec(),
```
Aeledfyr, line 270: This appears to be a double allocation.  The `upload_file` function should take a slice, rather than a `Vec<u8>`.

src/routes/mod_creation.rs
```diff
@@ -0,0 +1,401 @@
...
+                                            hashes: vec![FileHash {
+                                                algorithm: "sha1".to_string(),
+                                                hash: upload_data.content_sha1,
+                                            }],
+                                            url: format!("{}/{}", cdn_url, upload_data.file_name),
+                                        }],
+                                        dependencies: version_data
+                                            .dependencies
+                                            .into_iter()
+                                            .map(|x| x.0 as i32)
```
Aeledfyr, line 298: This is currently reinterpreting the input `ModId` (which should have been a `VersionId`) as a `VersionId`.  Be careful about that - in this place the other location is incorrect, but it could happen elsewhere.

src/routes/mod_creation.rs
```diff
@@ -0,0 +1,228 @@
...
+            icon_url,
+            description: create_data.mod_description,
+            body_url: format!("cdnurl/{}", body_url),
+            published: Utc::now(),
+            downloads: 0,
+            categories: create_data.categories,
+            versions: version_ids,
+            issues_url: create_data.issues_url,
+            source_url: create_data.source_url,
+            wiki_url: create_data.wiki_url,
```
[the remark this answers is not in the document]
Aeledfyr: Also, the **requires** validation, because otherwise people could potentially put in `javascript:` urls for XSS.

src/database/models/version_item.rs
Geometrically: The Version ID of the dependencies. I'll add docs

src/routes/mod_creation.rs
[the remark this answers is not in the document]
Geometrically: I didn't see this, mistake on my part. No idea why I put it in the Mod struct

src/routes/mod_creation.rs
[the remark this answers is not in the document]
Geometrically: I think that's out of scope of this PR. A seperate one should be created to address this.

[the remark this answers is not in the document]
Geometrically: #37

src/routes/mod_creation.rs
Geometrically: Good point - I'll do this real quick as a safety measure.

src/routes/mod_creation.rs
Geometrically: Will fix this.

src/routes/mod_creation.rs
Geometrically: Will fix this.

src/models/mods.rs
Geometrically: I re-added the field. I think it's a little too complex for modders if every single file had a loader. If it's a forge version, it should probably not be with the fabric version as that causes confusion. It would also make it harder to present downloads on the site.

Geometrically: The reason why loaders is a vector is that some mods have the forge version and fabric version bundled.

src/routes/mod_creation.rs
Geometrically: Okay, I figured out why I didn't do this earlier, reqwest doesn't take in a slice for it's body for some reason. (they do but only one with a static lifetime)

src/routes/mod_creation.rs
Geometrically: Okay, I figured out why I didn't do this earlier, reqwest doesn't take in a slice for it's body for some reason. (they do but only one with a static lifetime)

Aeledfyr: Hmm.  That's not optimal, but we should still be able to reduce this case to a single allocation rather than two.

src/main.rs
```diff
@@ -19,9 +19,28 @@ async fn main() -> std::io::Result<()> {
 
     check_env_vars();
 
+    //Database Connecter
```
Scotsguy, line 22: Connect*o*r
also, not gonna do this for every comment, but space after //

src/models/mods.rs
```diff
@@ -72,9 +73,11 @@ pub struct Version {
     /// A list of files available for download for this version.
     pub files: Vec<VersionFile>,
     /// A list of mods that this version depends on.
-    pub dependencies: Vec<ModId>,
+    pub dependencies: Vec<VersionId>,
     /// A list of versions of Minecraft that this version of the mod supports.
     pub game_versions: Vec<GameVersion>,
+    /// The loaders that this version works on
+    pub loaders: Vec<String>,
```
Scotsguy, line 80:
```suggestion
    pub loaders: Vec<ModLoader>,
```

PR #38, opened 2020-07-17
Create a mock file host for dev, Fix mod creation route

Aeledfyr:
This implements a mock file hosting system backed by the system's filesystem.  It mirrors the API of the backblaze integration, but puts the files directly on disk in the path specified by the `MOCK_FILE_PATH` environment variable (defaults to `/tmp/modrinth`).

The mock file hosting is enabled by default using cargo features to allow people to work on modrinth without access to a valid backblaze account and setup.  To enable backblaze, specify the cargo feature "backblaze" when running, ex. `cargo run --features backblaze`.

This PR also has multiple fixes for the mod creation route, which should now be functional.


Aeledfyr:
Whoops, didn't see that because of different feature set.  Is there any good way of making sure every combination of features compiles / passes clippy?

PR #39, opened 2020-07-19
Switch to Postgres

Scotsguy:


src/routes/mod_creation.rs
```diff
@@ -9,25 +9,22 @@ use actix_web::http::StatusCode;
 use actix_web::web::Data;
 use actix_web::{post, HttpResponse};
 use bson::doc;
```
Geometrically, line 11: Remove import so it compiles

src/routes/mod_creation.rs
```diff
@@ -327,17 +327,33 @@ pub async fn mod_create(
...
+            ",
+        )
+        .bind(version.version_id)
+        .bind(version.mod_id)
+        .bind(&version.name)
+        .bind(&version.version_number)
+        .bind(&version.changelog_url)
+        .bind(0) // TODO: release channel
+        .execute(client.as_ref())
+        .await?;
```
Geometrically, line 356: This should be after the mod is inserted into the database so it works.

src/routes/mod_creation.rs
Scotsguy: I thought I did
```suggestion

```

src/database/models/mod_item.rs
```diff
@@ -1,13 +1,10 @@
...
-use crate::database::models::Item;
-use crate::database::Result;
-use bson::{Bson, Document};
 use serde::{Deserialize, Serialize};
 
 #[derive(Deserialize, Serialize)]
 pub struct Mod {
     /// The ID for the mod, must be serializable to base62
-    pub id: i32,
+    pub id: i64,
```
Aeledfyr, line 7: We should create wrapper types for the different types of Ids, so that they don't get mixed up.  Also, the `version_ids` field below must use `i64` or a wrapped `i64`. (For some reason I can't comment on that line).

src/database/models/team_item.rs
```diff
@@ -4,7 +4,7 @@ use serde::{Deserialize, Serialize};
 #[derive(Serialize, Deserialize)]
 pub struct Team {
     /// The id of the team
-    pub id: i32,
+    pub id: i64,
     /// A list of the members of the team
     pub members: Vec<TeamMember>,
```
Aeledfyr, line 9: This should be a list of the ids of team members, since it is the database representation.  Although, since this is a representation of the database models, should it even be here?  It would require a second SQL query to retrieve, and doesn't match the structure of the teams table.  We should remove all of the `Vec`s in the database models, since those are left over from the mongodb model.

src/database/models/version_item.rs
```diff
@@ -1,30 +1,26 @@
...
-    pub number: String,
+    pub version_number: String,
     pub changelog_url: Option<String>,
     pub date_published: String,
     pub downloads: i32,
     pub files: Vec<VersionFile>,
-    pub dependencies: Vec<i32>,
+    pub dependencies: Vec<i64>,
     pub game_versions: Vec<String>,
     pub loaders: Vec<String>,
```
Aeledfyr, line 18: These don't match the database model for a version.  What's our plan for representing them here?

src/routes/mod_creation.rs
```diff
@@ -110,25 +104,33 @@ struct ModCreateData {
 #[post("api/v1/mod")]
 pub async fn mod_create(
     mut payload: Multipart,
-    client: Data<Client>,
+    client: Data<PgPool>,
```
Aeledfyr, line 107: We need to make a transaction here, and add error handling at each place that this could return an error to delete files that have been uploaded when the transaction fails.

src/search/indexing/local_import.rs
```diff
@@ -1,70 +1,108 @@
...
+                    INNER JOIN categories c ON mc.joining_category_id=c.id
+                WHERE mc.joining_mod_id = $1
+                ",
+                result.id
+            )
+            .fetch_all(&pool)
+            .await?
+            .into_iter()
+            .map(|c| c.category)
+            .collect::<Vec<String>>();
```
Aeledfyr, line 79: Is there a better way of doing this that doesn't require allocating two `Vec`s?  It seems like it should be possible using streams and stream combinators.

Redblueflame:
Seems like everything that is needed for the migration is done,,I'll mark this PR as ready to review

Aeledfyr:
The current remaining things to fix are:
- [ ]  Tracking cdn uploads that can't been deleted when a transaction fails (due to a cdn outage) - should be moved to a github issue & later PR
- [ ] Release channel lookups rather than fixed integer ids
- [ ] Add `game_version` and `loaders` info to mod versions
- [ ] Implement categories and convert the category strings to `CategoryId`s
- [ ] Figure out why sqlx broke CI this time

Redblueflame:
@Aeledfyr I think that everything other than the fist one is out of scope for this PR. We added back every functionality that were present on the current `master` branch.

Aeledfyr:
Then all we need is to fix CI again for it to be good to merge.  The first one is much more out of the scope of the PR than the rest of them, but these can all be added afterwards. 
It seems the CI is failing due to sqlx not having the proper data files, but `cargo sqlx prepare --check` works for me, and the data files have been committed.  On my machine, the build works with `SQLX_OFFLINE=true` without giving the errors that it does in github actions, so I don't know why it doesn't work.

Redblueflame:
Last build, with the modification from @Scotsguy, was working.

Redblueflame:
And there is still the ICE in place with `reqwest` in nightly, if we can merge the add docker to this branch, there is one less reason for everything to break.

PR #40, opened 2020-07-22
Add dockerfile

Redblueflame:
Adds dockerfile & automatic building of images
TODO:
- [x]  Add something to tag the latest build.


Cargo.toml
```diff
@@ -1,10 +1,14 @@
 [package]
-name = "modrinth"
+name = "labrinth"
 version = "0.1.0"
 #Team members, please add your emails and usernames
 authors = ["geometrically <jai.a@tuta.io>", "Redblueflame <contact@redblueflame.com>", "Aeledfyr <aeledfyr@gmail.com>", "cfanoulis"]
```
cfanoulis, line 5:
```suggestion
authors = ["geometrically <jai.a@tuta.io>", "Redblueflame <contact@redblueflame.com>", "Aeledfyr <aeledfyr@gmail.com>", "Charalampos Fanoulis <yo@fanoulis.dev>"]
```

Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM rust:1.43.1 as build
```
cfanoulis, line 1: Can we maybe get an alipne (=more lightweight) version? or is this one the lightest one

Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+# Build only deps
+RUN cargo build --release
+# Now return the file back to normal
+RUN sed -i 's|dummy.rs|src/main.rs|' Cargo.toml
+
+# Copy everything
+COPY . .
+# Build our code
+ARG SQLX_OFFLINE=true
+RUN cargo build --release
```
cfanoulis, line 20: why do we build twice? Just yeet the code in and then download dependencies

Dockerfile
Redblueflame: This is a multi stage build, and is only used for compilation.

Dockerfile
Redblueflame: It is made like this to allow for caching. The deps will only be donwloaded and compiled if there is a change in Cargo.toml, Cargo.lock or dummy.rs.
It can easily speed up build times.

Redblueflame:
And I broke everything by merging the most recent changes 😓

.github/workflows/tests.yml
```diff
@@ -19,7 +19,6 @@ jobs:
         rust:
           - stable
           - beta
-          - nightly
```
Scotsguy, line 22: I can't actually make this a suggestion (damn you, github!), but:
```yaml
jobs:
  test:
    name: ${{ matrix.os }}-rust-${{ matrix.rust }}
    runs-on: ${{ matrix.os }}
    continue-on-error: ${{ matrix.continue-on-error }}
    strategy:
      matrix:
        os: [ubuntu-latest]
        rust: [beta, nightly]
        continue-on-error: [false]
        include:
          - os: ubuntu-latest
          - rust: stable
          - continue-on-error: true

```

Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM rust:1.43.1 as build
+ENV PKG_CONFIG_ALLOW_CROSS=1
+
+WORKDIR /usr/src/labrinth
+# Download and compile deps
+COPY Cargo.toml .
+COPY Cargo.lock .
+COPY docker_utils/dummy.rs .
```
Scotsguy, line 8: maybe just `echo "fn main() {}" > dummy.rs`? One less useless file in git

Redblueflame: It's a possibility, but that means that we have to edit the docker file to fix a dependency compilation inconsistency, so break the whole caching system

PR #44, opened 2020-07-26
Refactor Meilisearch, update to latest SDK, and implement faceted search

### LINKED ISSUE #9: Switch to faceted search for filters
Faceted Search is a new feature added in MeiliSearch v0.11. It improves how filters work and increases performance for them with huge margins. 

It should be a simple addition, as the current tags system basically drops right into facets.

### LINKED ISSUE #17: CurseForge Projects currently have negative ids
This is an easy fix.

MeiliSearch currently parses integers as strings, so we could add a prefix to the id like "curse-2939291"

Aeledfyr:
This is currently using a fork of meilisearch with a few changes ([Aeledfyr/meilisearch-rust](https://github.com/Aeledfyr/meilisearch-rust)).  The main changes are with error handling to add support for Meilisearch v0.12.

This implements faceted search for mod categories and updates to the meilisearch master, which has async support.

This PR also implements a basic scheduler for reindexing mods at a given frequency, and indexes newly created mods every 15 minutes.  

This fixes #17 and may fix #9, depending on whether we should also use faceted search for filtering by version.


src/main.rs
```diff
@@ -44,22 +43,108 @@ async fn main() -> std::io::Result<()> {
...
+    //         .expect("Mod indexing failed");
+    // } else if !exe_path.exists() {
+    //     // The indexes were not created, or the version was upgraded
+    //     info!("Indexing of mods for first time...");
+    //     index_mods(pool.clone(), indexing_settings)
+    //         .await
+    //         .expect("Mod indexing failed");
+    //     // Create the lock file
+    //     File::create(exe_path)?;
+    // }
```
Geometrically, line 73: Could we async index so the server still starts while the indexing is being done?

src/search/indexing/local_import.rs
```diff
@@ -53,23 +53,23 @@ pub async fn index_local(pool: PgPool) -> Result<Vec<SearchMod>, IndexingError>
...
                 author_url: "".to_string(),
                 date_created: result.published.to_string(),
-                created: 0,
+                created_timestamp: 0,
                 date_modified: "".to_string(),
-                updated: 0,
+                modified_timestamp: 0,
                 latest_version: "".to_string(),
-                empty: String::from("{}{}{}"),
+                empty: std::borrow::Cow::Borrowed("{}{}{}"),
```
Geometrically, line 72: Might be out of scope of this PR, but we should fill in these values with proper stuff since a lot of the database stuff has been done.

src/main.rs
Aeledfyr: That's what the system currently does.  The scheduler triggers the indexing once on its own threadpool/arbiter when the server is started, so the server starts while the indexing is being processed.  The main question with this part of the code is whether we want an option to not index at startup, so that rapid iteration and other development is easier without the indexing spamming the logs.
We could set up an environment variable to control whether the server indexes on startup and then every hour (the configured interval), or every hour after startup, but not at startup.

src/search/indexing/local_import.rs
Aeledfyr: The only things that could change here without much work would be adding the `created_timestamp` as `result.published.timestamp()`.  We don't currently store the last modified time in the database for mods, we have no direct connection between mods and users for the author - we may be able to replace this with teams, and we don't currently have a system for tracking the latest version (This will likely require a separate PR, because it is complicated and is going to be hard to get right.)

Technically we could return the update time and timestamp from the created time and timestamp, but at some point we will need to change that.

src/main.rs
Aeledfyr: I've used the argument system from before, but with `skip` instead of `regen` to allow the user to skip the initial mod indexing, which should cover what this code did before, besides checking the cache.

src/search/indexing/local_import.rs
Aeledfyr: For now I've added the modified times as the created time, but the author/latest version fields are still blank.  I'll leave the rest for a later PR.

src/main.rs
Geometrically: I think it's fine the current way where it does it asynchronously. Is there a log output every time it indexes?

PR #47, opened 2020-07-31
Version Creation

Geometrically:
A simple PR for version creation and adding responses to the creation routes. Stuff left todo:

- [ ] Tests
- [x] File creation (Uploading a file to an existing version)
- [x] Making sure a duplicate version identifier isn't uploaded

### DIFF
--- a/src/routes/mod.rs
+++ b/src/routes/mod.rs
@@ -8,3 +8,5 @@ pub use self::index::index_get;
 pub use self::mod_creation::mod_create;
 pub use self::mods::mod_search;
 pub use self::not_found::not_found;
+pub use self::version_creation::upload_file_to_version;
+pub use self::version_creation::version_create;
--- a/src/routes/mod.rs
+++ b/src/routes/mod.rs
@@ -1,7 +1,7 @@
 mod index;
+mod mod_creation;
 mod mods;
 mod not_found;
-mod mod_creation;
 mod version_creation;
 
 pub use self::index::index_get;
--- a/src/routes/mod.rs
+++ b/src/routes/mod.rs
@@ -1,7 +1,8 @@
 mod index;
-mod mod_creation;
 mod mods;
 mod not_found;
+mod mod_creation;
+mod version_creation;
 
 pub use self::index::index_get;
 pub use self::mod_creation::mod_create;
--- a/src/routes/mod_creation.rs
+++ b/src/routes/mod_creation.rs
@@ -13,6 +13,7 @@ use serde::{Deserialize, Serialize};
 use sqlx::postgres::PgPool;
 use std::sync::Arc;
 use thiserror::Error;
+use crate::routes::version_creation::InitialVersionData;
 
 #[derive(Error, Debug)]
 pub enum CreateError {
@@ -70,18 +71,6 @@ impl actix_web::ResponseError for CreateError {
 }
 
 #[derive(Serialize, Deserialize, Clone)]
-struct InitialVersionData {
-    pub file_parts: Vec<String>,
-    pub version_number: String,
-    pub version_title: String,
-    pub version_body: String,
-    pub dependencies: Vec<VersionId>,
-    pub game_versions: Vec<GameVersion>,
-    pub release_channel: VersionType,
-    pub loaders: Vec<String>,
-}
-
-#[derive(Serialize, Deserialize, Clone)]
 struct ModCreateData {
     /// The title or name of the mod.
     pub mod_name: String,
@@ -105,9 +94,9 @@ struct ModCreateData {
     pub wiki_url: Option<String>,
 }
 
-struct UploadedFile {
-    file_id: String,
-    file_name: String,
+pub struct UploadedFile {
+    pub file_id: String,
+    pub file_name: String,
 }
 
 async fn undo_uploads(
--- a/src/routes/mod_creation.rs
+++ b/src/routes/mod_creation.rs
@@ -419,7 +419,11 @@ async fn mod_create_inner(
         published: now,
         downloads: 0,
         categories: create_data.categories.clone(),
-        versions: mod_builder.initial_versions.iter().map(|v| crate::models::ids::VersionId(v.version_id.0 as u64)).collect::<Vec<_>>(),
+        versions: mod_builder
+            .initial_versions
+            .iter()
+            .map(|v| crate::models::ids::VersionId(v.version_id.0 as u64))
+            .collect::<Vec<_>>(),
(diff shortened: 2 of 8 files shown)


src/routes/mod.rs
```diff
@@ -2,6 +2,7 @@ mod index;
 mod mod_creation;
 mod mods;
 mod not_found;
+mod version_creation;
```
Aeledfyr, line 5: Don't forget to re-export the route and add it to the actix app.

src/routes/mod_creation.rs
```diff
@@ -421,10 +410,30 @@ async fn mod_create_inner(
 
     indexing_queue.add(index_mod);
 
+    let response = crate::models::mods::Mod {
+        id: mod_id,
+        team: crate::models::ids::TeamId(team_id.0 as u64),
```
Aeledfyr, line 415: You can use `team_id.into()` for this.

src/routes/mod_creation.rs
```diff
@@ -421,10 +410,30 @@ async fn mod_create_inner(
...
+        title: mod_builder.title.clone(),
+        description: mod_builder.description.clone(),
+        body_url: mod_builder.body_url.clone(),
+        published: now,
+        downloads: 0,
+        categories: create_data.categories.clone(),
+        versions: mod_builder
+            .initial_versions
+            .iter()
+            .map(|v| crate::models::ids::VersionId(v.version_id.0 as u64))
```
Aeledfyr, line 425: Same as above, `v.version_id.into()` works here. (most of the database Ids implement `Into` for the API model ids.)

src/routes/version_creation.rs
```diff
@@ -0,0 +1,239 @@
...
+            .ok_or_else(|| CreateError::MissingValueError("Missing content name".to_string()))?;
+
+        if name == "data" {
+            let mut data = Vec::new();
+            while let Some(chunk) = field.next().await {
+                data.extend_from_slice(&chunk.map_err(CreateError::MultipartError)?);
+            }
+
+            let version_create_data: InitialVersionData = serde_json::from_slice(&data)?;
+            initial_version_data = Some(version_create_data.clone());
```
Aeledfyr, line 90: This clone seems unnecessary.  You may be able to avoid the extra allocations by making `version_create_data` a reference to the value that has been moved into the option.
```rust
let version_create_data: InitialVersionData = serde_json::from_slice(&data)?;
initial_version_data = Some(version_create_data);
let version_create_data = initial_version_data.as_ref().unwrap();
```

src/routes/version_creation.rs
```diff
@@ -0,0 +1,239 @@
...
+            })?;
+
+            let results = sqlx::query!(
+                "SELECT EXISTS(SELECT 1 FROM mods WHERE id=$1)",
+                mod_id.0 as i64
+            )
+            .fetch_one(&mut *transaction)
+            .await?;
+
+            if !results.exists.unwrap_or(true) {
```
Aeledfyr, line 103: This should be `.unwrap_or(false)`, since the unwrapped value is immediately inverted.  If the query fails (I don't think it can?) it should act like it didn't find any mods for that id.

src/routes/version_creation.rs
```diff
@@ -0,0 +1,239 @@
...
+            uploaded_files.push(UploadedFile {
+                file_id: uploaded_text.file_id.clone(),
+                file_name: uploaded_text.file_name.clone(),
+            });
+
+            // TODO: do a real lookup for the channels
+            let release_channel = match version_create_data.release_channel {
+                VersionType::Release => models::ChannelId(0),
+                VersionType::Beta => models::ChannelId(2),
+                VersionType::Alpha => models::ChannelId(4),
```
Aeledfyr, line 129: I made a minor change to the version that this was copied from to fix the sqlx errors.  The release channel ids start at 1, so `Release` should be 1, `Beta` should be 3, and `Alpha` should be 5.

src/routes/version_creation.rs
```diff
@@ -0,0 +1,239 @@
...
+            };
+
+            version_builder = Some(VersionBuilder {
+                version_id: version_id.into(),
+                mod_id: models::ModId(mod_id.0 as i64),
```
Aeledfyr, line 134: This should be `mod_id.into()`.

src/routes/version_creation.rs
```diff
@@ -0,0 +1,239 @@
...
+            while let Some(chunk) = field.next().await {
+                data.extend_from_slice(&chunk.map_err(CreateError::MultipartError)?);
+            }
+
+            let upload_data = file_host
+                .upload_file(
+                    "application/java-archive",
+                    &format!(
+                        "{}/{}/{}",
+                        version.mod_id.0, version.version_number, file_name
```
Aeledfyr, line 179: This should be just `version.mod_id` so that it is formatted as base62.  While using just the integer works, it isn't consistent with the rest of the ids.

src/routes/version_creation.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    let version_data_safe = initial_version_data
+        .ok_or_else(|| CreateError::InvalidInput("`data` field is required".to_string()))?;
+    let version_builder_safe = version_builder
+        .ok_or_else(|| CreateError::InvalidInput("`data` field is required".to_string()))?;
+
+    let response = Version {
+        id: VersionId(version_builder_safe.version_id.0 as u64),
+        mod_id: ModId(version_builder_safe.mod_id.0 as u64),
```
Aeledfyr, line 215: Could use `.into()`.

src/routes/version_creation.rs
```diff
@@ -0,0 +1,239 @@
...
+        downloads: 0,
+        version_type: version_data_safe.release_channel,
+        files: version_builder_safe
+            .files
+            .iter()
+            .map(|_x| VersionFile {
+                // TODO: Put real data here
+                hashes: vec![],
+                url: String::new(),
+            })
```
Aeledfyr, line 229: I believe this has all of the data available if this is returning `VersionFile`s instead of `FileId`s:
```rust
.map(|file| VersionFile {
    hashes: file.hashes.iter().map(|hash| FileHash {
        algorithm: hash.algorithm,
        // This is a hack since the hashes are currently stored as ASCII
        // in the database, but represented here as a Vec<u8>.  At some
        // point we need to change the hash to be the real bytes  in the
        // database and add more processing here.
        hash: String::from_utf8(hash.hash).unwrap(),
    }).collect(),
    url: file.url,
})
```

PR #50, opened 2020-08-12
Implement more database methods and basic API routes

Aeledfyr:
This significantly improves the API in `database` module, and implements some new routes for the external API.

This also adds a way of managing `loaders`, `categories` and `game_versions` through the API under the `/tag` route group.

The API routes have been reorganized in a hopefully modular way by using actix's `scope`s and `.configure` methods on the service constructors.

Here is the current definition of most of the routes - it's not great, but it's better than putting it all in the `main.rs` file linearly.
```rust
pub fn mods_config(cfg: &mut web::ServiceConfig) {
    cfg.service(mods::mod_search);
    cfg.service(mod_creation::mod_create);

    cfg.service(
        web::scope("mod")
            .service(mods::mod_get)
            .service(mods::mod_delete)
            .service(web::scope("{mod_id}").configure(versions_config)),
    );
}

pub fn versions_config(cfg: &mut web::ServiceConfig) {
    cfg.service(versions::version_list)
        .service(version_creation::version_create)
        .service(
            web::scope("version")
                .service(versions::version_get)
                .service(versions::version_delete)
                .service(
                    web::scope("{version_id}").service(version_creation::upload_file_to_version),
                ),
        );
}
```


Aeledfyr:
Nope, you can still add middleware by using the `.wrap` method on base `App`, or by adding it to specific scopes.  This method actually gives better control for the middleware, because it can limit specific middleware to a smaller scope of routes.

Geometrically:
Sounds good! This looks great by the way - and will make development and working with sqlx a lot easier.

PR #54, opened 2020-08-28
Added automatic migration for the database.

Redblueflame:


.env
```diff
@@ -3,11 +3,10 @@ RUST_LOG=info,sqlx::query=warn
 
 CDN_URL=https://cdn.modrinth.com
 
-DATABASE_URL=postgresql://labrinth@localhost/labrinth
+DATABASE_URL=postgresql://labrinth:labrinth@localhost/labrinth
```
Scotsguy, line 6: why?

Redblueflame: In the latest builds of postgres, removing the password for the user spits warnings in the logs, it's reflected in the docker-compose.yml file

Scotsguy: yeah, of course it does, it's set up to accept connections without authentication. that's intentional

PR #61, opened 2020-09-07
Allow for API user to change the amount of mods responded with in search

Geometrically:
Maximum of 100, to prevent abuse

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,4 +1,3 @@
-
 use std::fs;
 use std::path::{Path, PathBuf};
 fn main() {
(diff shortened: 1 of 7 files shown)


Aeledfyr:
This should be rebased/merged with master.  (Which should make reviewing changes easier)

Aeledfyr (approved):
I don't know what max limit is the best idea - 100 seems a bit large, but it could be fine.  We may want to test it with our Meilisearch backend to figure out what limit works best.
Other than that, we may want to either use an environment variable or a constant somewhere so that it's easy to find and change the max count.
This seems fine though, if you don't want to change anything.

PR #64, opened 2020-09-29
Make scopes safe for browser

Geometrically:


Redblueflame (approved):
Ok tier, but I'ld still remove the org read, and only keep the email (principle of least privileges).

sschr15 (approved):
looks good

PR #65, opened 2020-10-01
General cleanup: fix some bugs, some refactoring

Aeledfyr:
Notable changes:
* Merged mod file upload in version creation, mod creation and version file add to one function;  This makes sure that they are consistent
* Made some fields on `User` optional: `github_id`, `avatar_url`, `bio`.
    * We may not want to publicly show the `github_id` to everyone with access to the API
    * If we allow non-github users, some of those fields would be invalid; some oauth providers may not have avatars or bios
* Made CORS origins configurable
* Made `--reconfigure-indices` and `--reset-indices` exit after completion instead of starting the server

### DIFF
--- a/.env
+++ b/.env
@@ -1,6 +1,7 @@
 DEBUG=true
 RUST_LOG=info,sqlx::query=warn
 
+CORS_ORIGINS='["http://localhost:3000","https://modrinth.com"]'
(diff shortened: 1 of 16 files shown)


Geometrically (approved):
LGTM

PR #66, opened 2020-10-01
Add documentation

Aeledfyr:
This adds basic documentation for the environment variables, command line arguments, and API routes.

This still needs some work and verification, but it seems like a reasonable starting place.


Geometrically:
Looks great so far! However, do you think it would be better to make a swagger thing and move docs to another repo?

Aeledfyr:
For now, this is probably better;  We've been saying that we'll make proper documentation for a month but still haven't done anything because of the work it takes to set up and learn.  For the official API, once it's public, we should use swagger or a proper API documentation generator, but for now it's just better to have some docs rather than no docs.

Also, some of this documentation doesn't fit within the API docs, since this also partially documents the config for running the backend.  We could make some sort of hybrid documentation repo, but the documentation for config and setup should probably stay in this repo.

Geometrically:
Moved to the Wiki

PR #68, opened 2020-10-03
Add default bio value, to fix GitHub integration errors

Geometrically:


Aeledfyr:
Bio is optional in all of the other structures; just pass it through and don't add an default

Aeledfyr (approved):

PR #69, opened 2020-10-03
Fix another GitHub OAuth Bug, allow users to register with null names.

Geometrically:


migrations/20201003211651_make-name-null.sql
```diff
@@ -0,0 +1,5 @@
+-- Add migration script here
+ALTER TABLE users
+ALTER COLUMN name DROP NOT NULL;
+ALTER TABLE users
+ALTER COLUMN name DROP DEFAULT;
```
Aeledfyr, line 5: I already removed the default in my bug-fix PR, it may cause issues if done twice.

Aeledfyr (approved):
Make sure to run `cargo sqlx prepare`, but other than that it looks good.  Also check that dropping the default twice won't cause issues.

migrations/20201003211651_make-name-null.sql
Geometrically: It worked fine for me

PR #70, opened 2020-10-04
Add API routes to request multiple of an item

Geometrically:


src/database/models/user_item.rs
```diff
@@ -112,4 +113,45 @@ impl User {
...
+                ",
+                team_id as TeamId,
+            )
+                .fetch_many(exec)
+                .try_filter_map(|e| async { Ok(e.right().map(|v| ModId(v.id))) })
+                .try_collect::<Vec<ModId>>()
+                .await?;
+
+            mods.append(&mut vec);
+        }
```
Aeledfyr, line 152: This should be able to be done with one query, rather than (generally) `1 + number of mods` queries

src/routes/mod.rs
```diff
@@ -18,8 +18,15 @@ pub use self::not_found::not_found;
...
 pub fn mods_config(cfg: &mut web::ServiceConfig) {
     cfg.service(mods::mod_search);
+    cfg.service(mods::mods_get);
     cfg.service(mod_creation::mod_create);
 
+    cfg.service(versions::versions_get);
+    cfg.service(web::scope("version")
+        .service(versions::version_get)
+        .service(versions::version_delete)
+        .service(version_creation::upload_file_to_version));
```
Aeledfyr, line 28: This should be in `versions_config`

Aeledfyr (changes requested):
One, fix the failing tests & run `cargo fmt`.

I'm not sure about moving the `version` routes to the api root;  It makes sense, but then it makes less sense to have `mod/{mod_id}/versions` return a populated list of versions rather than a list of ids.

The routes for getting a list of entries seem useful, but is the naming the best?  It may become confusing to have both `/api/v1/mod` and `/api/v1/mods` as routes, one that searches and one that requests by id.

Geometrically:
> One, fix the failing checks & run `cargo fmt`.
> 
> I'm not sure about moving the `version` routes to the api root; It makes sense, but then it makes less sense to have `mod/{mod_id}/versions` return a populated list of versions rather than a list of ids.
> 
> The routes for getting a list of entries seem useful, but is the naming the best? It may become confusing to have both `/api/v1/mod` and `/api/v1/mods` as routes, one that searches and one that requests by id.

Fixed failing chekcs

The reason why it won't return all the versions is because we have the get multiple versions route, so the user can decide what versions they actually need/want.

It might be confusing, but that is what documentation is for. I don't know a better way of doing it

src/routes/mod.rs
```diff
@@ -21,23 +21,26 @@ pub fn mods_config(cfg: &mut web::ServiceConfig) {
...
-
     cfg.service(
         web::scope("mod")
             .service(mods::mod_get)
             .service(mods::mod_delete)
-            .service(web::scope("{mod_id}").configure(versions_config)),
+            .service(
+                web::scope("{mod_id}")
+                    .service(versions::version_list)
+                    .service(version_creation::version_create),
```
Aeledfyr, line 31: Why is creating a version under `/api/v1/mod/{mod_id}/version`, while everything else with versions is in `/api/v1/version`?  Wouldn't it make more sense to keep them together, since I believe that the version creation request includes the mod id?

src/routes/mods.rs
```diff
@@ -15,13 +16,49 @@ pub async fn mod_search(
...
+) -> Result<HttpResponse, ApiError> {
+    let mut mods = vec![];
+
+    for id in serde_json::from_str::<Vec<models::ids::ModId>>(&*ids.ids)? {
+        let mod_data = get_mod_from_id(id, &*pool).await?;
+
+        if let Some(data) = mod_data {
+            mods.push(data)
+        }
+    }
```
Aeledfyr, line 37: This is inefficient, since it makes a database query for every requested mod.  It appears that it is possible to use `WHERE id IN (...)` using `sqlx`: (https://github.com/launchbadge/sqlx/issues/656#issuecomment-699638262); this would make it more efficient, and make it less likely to break our servers in case someone tries to do what we're doing with curseforge.

src/routes/users.rs
```diff
@@ -22,13 +24,49 @@ pub async fn user_auth_get(
...
+) -> Result<HttpResponse, ApiError> {
+    let mut users = vec![];
+
+    for id in serde_json::from_str::<Vec<UserId>>(&*ids.ids)? {
+        let mod_data = get_user_from_id(id, &*pool).await?;
+
+        if let Some(data) = mod_data {
+            users.push(data)
+        }
+    }
```
Aeledfyr, line 45: Same as for the `mods_get` comment;  We could either have a local function that implements it more efficiently or add implement it in the `database` module.

src/routes/versions.rs
```diff
@@ -42,24 +43,55 @@ pub async fn version_list(
...
+) -> Result<HttpResponse, ApiError> {
+    let mut versions = vec![];
+
+    for id in serde_json::from_str::<Vec<models::ids::VersionId>>(&*ids.ids)? {
+        let mod_data = get_version_from_id(id, &*pool).await?;
+
+        if let Some(data) = mod_data {
+            versions.push(data)
+        }
+    }
```
Aeledfyr, line 64: Same as other comments above on efficiency

Aeledfyr (changes requested):
Batched database queries should be an order of magnitude faster, so we should at least try them.  If it doesn't work out then what we have may be good enough, but it seems like it should be easy enough to implement.

src/database/models/mod_item.rs
```diff
@@ -143,6 +143,45 @@ impl Mod {
         }
     }
 
+    pub async fn get_many<'a, E>(mod_ids: Vec<i64>, exec: E) -> Result<Vec<Mod>, sqlx::Error>
```
Aeledfyr, line 146: These methods should still take `Vec<ModId>` or `Vec<UserId>`.  That may just work with the database (you'd have to try it) but the best way would be to convert it within the method, rather than trusting that the caller used the right id type.

src/routes/mods.rs
```diff
@@ -15,6 +16,45 @@ pub async fn mod_search(
...
+}
+
+// TODO: Make this return the full mod struct
+#[get("mods")]
+pub async fn mods_get(
+    web::Query(ids): web::Query<ModIds>,
+    pool: web::Data<PgPool>,
+) -> Result<HttpResponse, ApiError> {
+    let mods_data =
+        database::models::Mod::get_many(serde_json::from_str::<Vec<i64>>(&*ids.ids)?, &**pool)
```
Aeledfyr, line 31: This is incorrect;  The input for this should be a JSON array of mod ids represented as base62 strings.  (We could use integers, but we need to be consistent, and we use base62 everywhere else).

This also applies to the other "get multiple" routes; they should all take base62 ids.

src/routes/mods.rs
```diff
@@ -15,6 +16,45 @@ pub async fn mod_search(
     Ok(HttpResponse::Ok().json(results))
 }
 
+#[derive(Serialize, Deserialize)]
+pub struct ModIds {
+    pub ids: String,
+}
+
+// TODO: Make this return the full mod struct
```
Aeledfyr, line 24: This TODO seems to be completed

src/routes/users.rs
```diff
@@ -50,6 +84,38 @@ pub async fn user_get(
     }
 }
 
+#[get("{id}/mods")]
+pub async fn mods_list(
```
Aeledfyr, line 88: A name like `user_mods_list` may be better for this, or even just change the `{id}` to `{user_id}` to make what this route does more obvious.

Aeledfyr (changes requested):
Still a few things left, but it's almost done.  Now it's mostly just the array types for the batch get routes and database methods.

src/routes/mods.rs
Geometrically: It isn't as the get multiple mods returns part of the mod, not the loaders/categories

Aeledfyr (approved):

PR #71, opened 2020-10-11
Fix date format

Geometrically:


Aeledfyr:
If you change the format here, you also have to change the format in the local database importer.

Scotsguy (approved):

Geometrically:
The database is already in RFC 3339

Aeledfyr:
The database is, but the indexer for the local database uses the same format that this originally did.  You have to change the other indexer to use `to_rfc3339` instead of `to_string`, or local mods will have a different format from external mods.

PR #73, opened 2020-10-13
Lots of little fixes

Geometrically:
This PR fixes some TODOs and removes redundant ones, adds some fields to the `Mod` Struct, and much more!


src/routes/mods.rs
```diff
@@ -78,6 +86,14 @@ pub async fn mod_get(
             description: m.description,
             body_url: m.body_url,
             published: m.published,
+            updated: m.updated,
+            status: match m.status.0 {
```
Aeledfyr, line 90: Same as the above.  Also, this is a significant amount of duplication; we should move the `database::models::Mod` to `models::Mod` conversion to a separate function that we can reuse.

src/routes/tags.rs
```diff
@@ -20,17 +20,12 @@ pub fn config(cfg: &mut web::ServiceConfig) {
     );
 }
 
-// TODO: searching / filtering? Could be used to implement a live
-// searching category list
```
Aeledfyr, line 24: Same with all of these;  Don't remove TODOs that aren't fixed yet.  This one may not be that useful, but you should justify how each of these are fixed or not needed.

src/routes/tags.rs
```diff
@@ -20,17 +20,12 @@ pub fn config(cfg: &mut web::ServiceConfig) {
...
 #[get("category")]
 pub async fn category_list(pool: web::Data<PgPool>) -> Result<HttpResponse, ApiError> {
     let results = Category::list(&**pool).await?;
     Ok(HttpResponse::Ok().json(results))
 }
 
-// At some point this may take more info, but it should be able to
-// remain idempotent
-// TODO: don't fail if category already exists
```
Aeledfyr, line 33: This still fails if the category already exists.  This should also still have more info, such as a category description.

src/routes/version_creation.rs
```diff
@@ -167,13 +167,28 @@ async fn version_create_inner(
                 file_name: uploaded_text.file_name.clone(),
             });
 
-            // TODO: do a real lookup for the channels
             let release_channel = match version_create_data.release_channel {
                 VersionType::Release => models::ChannelId(1),
                 VersionType::Beta => models::ChannelId(3),
                 VersionType::Alpha => models::ChannelId(5),
             };
```
Aeledfyr, line 174: Still an issue.  This should probably be shared between version creation and mod creation so that they are consistent.

src/search/indexing/local_import.rs
```diff
@@ -29,12 +28,11 @@ pub async fn index_local(pool: PgPool) -> Result<Vec<UploadSearchMod>, IndexingE
...
                 result.id
             )
-            .fetch_many(&pool)
-            .try_filter_map(|e| async { Ok(e.right().map(|c| c.version)) })
-            .try_collect::<Vec<String>>()
-            .await?;
+                .fetch_many(&pool)
+                .try_filter_map(|e| async { Ok(e.right().map(|c| c.version)) })
+                .try_collect::<Vec<String>>()
+                .await?;
```
Aeledfyr, line 34: Don't forget to run `rustfmt`

src/search/indexing/local_import.rs
```diff
@@ -29,12 +28,11 @@ pub async fn index_local(pool: PgPool) -> Result<Vec<UploadSearchMod>, IndexingE
...
 
-            // TODO: only loaders for recent versions? For mods that have moved from forge to fabric
```
Aeledfyr, line 37: This is still an issue, although it may not be the best place for this.  We may want to move these to github issues or a centralized list.

src/search/indexing/local_import.rs
```diff
@@ -58,20 +56,30 @@ pub async fn index_local(pool: PgPool) -> Result<Vec<UploadSearchMod>, IndexingE
...
+                .try_filter_map(|e| async { Ok(e.right().map(|c| c.category)) })
+                .try_collect::<Vec<String>>()
+                .await?;
 
             categories.extend(loaders);
 
+            let user = sqlx::query!(
+                "
+                SELECT u.id, u.username FROM users u
+                INNER JOIN team_members tm ON tm.role = 'Owner'
```
Aeledfyr, line 69: Don't hardcode the role names, or at least define a system for roles.  Whatever it is, we will want to be consistent with the naming convention.
`Author`, `author`, `Owner`, `owner`, `creator`, etc...
Do we want to use the original creator, or the current owner of the mod?

src/search/indexing/local_import.rs
```diff
@@ -58,20 +56,30 @@ pub async fn index_local(pool: PgPool) -> Result<Vec<UploadSearchMod>, IndexingE
...
+                .await?;
+
             let mut icon_url = "".to_string();
 
             if let Some(url) = result.icon_url {
                 icon_url = url;
             }
 
-            let formatted = result.published.to_rfc3339();
             let timestamp = result.published.timestamp();
```
Aeledfyr, line 83: Remove this, it's been moved into the struct constructor.

src/search/indexing/mod.rs
```diff
@@ -155,8 +155,13 @@ async fn create_index<'a>(
...
     match client.get_index(name).await {
-        // TODO: update index settings on startup (or delete old indices on startup)
-        Ok(index) => Ok(index),
+        Ok(index) => {
+            index
+                .set_settings(&default_settings().with_ranking_rules(rules()))
+                .await?;
+
+            Ok(index)
+        },
```
Aeledfyr, line 164: This shouldn't be done every time the method is called.  This method is called for every index every time a batch of mods is added to the index;  updating the settings constantly seems like a bad idea and may be very inefficient.
This case is already covered by the `--reconfigure-indices` commandline option; if we need to, it can be enabled by default.

src/routes/mod_creation.rs
```diff
@@ -394,6 +394,7 @@ async fn mod_create_inner(
 
         categories,
         initial_versions: created_versions,
+        status: models::ids::StatusId(1)
```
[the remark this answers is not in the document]
Aeledfyr: Looking up a status should be easy: `SELECT id FROM statuses WHERE status = $1`

src/search/indexing/local_import.rs
Aeledfyr: This could just be an enum for now, like `Status`.  It's just better not to hardcode the naming into the SQL.

Aeledfyr (changes requested):
One major point is that we need to organize the TODOs as although some of them have been fixed by this, many of the ones that were removed should either just be moved to a different place or haven't been fixed at all.

For example, removing the TODO for the modified datetime is technically correct, but we still haven't implemented updating the modified date for a mod.
Also, most of the TODOs in the tags area still need to be fixed.

Some of these may be able to be moved to github issues, but that seems too large for more minor unimplemented things.

src/search/indexing/local_import.rs
Geometrically: The owner name will always be hardcoded, as it's the only thing the user can't change. All other roles for other users are dynamic, however we need distinction to see who is owner of a mod.

src/search/indexing/mod.rs
Geometrically: Reverted.

src/routes/tags.rs
Geometrically: Isn't that the point? What should it do if the category exists

migrations/20201013183754_mod-statuses.sql
```diff
@@ -0,0 +1,10 @@
+-- Add migration script here
+CREATE TABLE statuses (
+    id bigint PRIMARY KEY,
+    status varchar(500)
+);
```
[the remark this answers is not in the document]
Geometrically: Added docs to the `ModStatus` struct, changed types.

src/database/models/ids.rs
```diff
@@ -102,9 +102,6 @@ pub struct TeamMemberId(pub i64);
 #[derive(Copy, Clone, Debug, Type)]
 #[sqlx(transparent)]
 pub struct ModId(pub i64);
-#[derive(Copy, Clone, Debug, Type)]
-#[sqlx(transparent)]
-pub struct StatusId(pub i64);
```
Aeledfyr: This should be kept, but changed to contain an `i32`.  This is in the database models, so it should represent the id in the database.

src/routes/mods.rs
```diff
@@ -46,7 +46,7 @@ pub async fn mods_get(
             body_url: m.body_url,
             published: m.published,
             updated: m.updated,
-            status: match m.status.0 {
+            status: match m.status {
```
Aeledfyr, line 49: This is why I recommended adding a more robust system for this.  This currently treats any `Processing` mods as `Rejected`, which is incorrect.

src/routes/tags.rs
```diff
@@ -93,9 +93,6 @@ pub async fn loader_list(pool: web::Data<PgPool>) -> Result<HttpResponse, ApiErr
     Ok(HttpResponse::Ok().json(results))
 }
 
-// At some point this may take more info, but it should be able to
-// remain idempotent
-// TODO: don't fail if loader already exists
```
Aeledfyr, line 98: Still an issue (Although I didn't put a comment directly on this one, it was still mentioned in the previous comments).

src/routes/tags.rs
```diff
@@ -158,8 +155,6 @@ pub async fn game_version_list(pool: web::Data<PgPool>) -> Result<HttpResponse,
     Ok(HttpResponse::Ok().json(results))
 }
 
-// At some point this may take more info, but it should be able to
-// remain idempotent
```
Aeledfyr, line 162: Same as the above, this is still an issue.

src/search/indexing/local_import.rs
Aeledfyr: If that will be true, then document it somewhere.  It is still a bad idea to inline the owner role into the SQL rather than using a rust constant or an enum.

migrations/20201013183754_mod-statuses.sql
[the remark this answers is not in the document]
Aeledfyr: 1. The `StatusId` wrapper should stay, and instead contain `i32`s.  Ids don't have to be publicly exposed to have to be representable internally.
2. Is there any reason for id being `integer` rather than `serial`? `serial` would make it autoincrement so that you don't have to specify the id of every status you insert.
3. Status is still limited to 500 characters.  Is there any reason for this?

Also, adding more migrations for this one PR makes it much less readable.  At the end, we should be able to merge them all into one migration, since none of this has been run on the main database.

Aeledfyr (changes requested):

src/routes/tags.rs
Aeledfyr: This method should act the same as when a category was created.  Either return a 200 or 204 status code, and do nothing.

migrations/20201014165954_create-statuses.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE statuses (
+    id serial PRIMARY KEY UNIQUE NOT NULL,
+    status varchar(64)
```
Aeledfyr, line 3: Might be good to make this `UNIQUE NOT NULL` too.

src/routes/mods.rs
```diff
@@ -36,23 +36,30 @@ pub async fn mods_get(
...
+                SELECT status FROM statuses
+                WHERE id = $1
+                ",
+                m.status.0,
+            )
+            .fetch_one(&**pool)
+            .await
+            .map_err(|e| ApiError::DatabaseError(e.into()))?
+            .status
+            .unwrap_or(ModStatus::Unknown.to_string());
```
Aeledfyr, line 52: This should be
```rust
.map(|r| models::mods::ModStatus::from_str(&r.status))
.unwrap_or(models::mods::ModStatus::Unknown)
```
to remove the unnecessary allocation and parsing.

src/routes/mods.rs
```diff
@@ -79,6 +86,19 @@ pub async fn mod_get(
...
+                SELECT status FROM statuses
+                WHERE id = $1
+                ",
+            m.status.0,
+        )
+        .fetch_one(&**pool)
+        .await
+        .map_err(|e| ApiError::DatabaseError(e.into()))?
+        .status
+        .unwrap_or(ModStatus::Unknown.to_string());
```
Aeledfyr, line 101: Same as the above.  This could also be improved by using `unwrap_or_else`, but that would still be less efficient than just converting it to a `ModStatus` here.

Aeledfyr (changes requested):
This seems to be almost done.  A few requests left: readd the last few TODOs in the tags routes, make `status` `UNIQUE NOT NULL`, and fix up a bit of the code.
I didn't specifically comment on it, but there are a few places where the formatting and indentation for some SQL queries are messed up; rustfmt can't change indentation within strings, so that has to be fixed manually.

src/routes/mod_creation.rs
```diff
@@ -267,12 +269,19 @@ async fn mod_create_inner(
...
-                VersionType::Beta => models::ChannelId(3),
-                VersionType::Alpha => models::ChannelId(5),
-            };
+            let release_channel = models::ChannelId(
+                sqlx::query!(
+                    "
+                SELECT id
+                FROM release_channels
+                WHERE channel = $1
+                ",
```
Aeledfyr, line 278: Formatting?

src/routes/mod_creation.rs
```diff
@@ -379,7 +388,18 @@ async fn mod_create_inner(
...
     let team_id = team.insert(&mut *transaction).await?;
 
-    // Insert the new mod into the database
+    let status = ModStatus::Processing;
+    let status_id = sqlx::query!(
+        "
+                SELECT id
+                FROM statuses
+                WHERE status = $1
+                ",
```
Aeledfyr, line 397: Formatting

src/routes/tags.rs
```diff
@@ -158,8 +158,6 @@ pub async fn game_version_list(pool: web::Data<PgPool>) -> Result<HttpResponse,
...
```
Aeledfyr, line 162: Re-add this, please.

src/routes/mods.rs
```diff
@@ -71,13 +85,28 @@ pub async fn mod_get(
...
+        let status = sqlx::query!(
+            "
+            SELECT status FROM statuses
+            WHERE id = $1
+            ",
+            m.status.0,
+        )
+        .fetch_one(&**pool)
+        .await
+            .map_err(|e| ApiError::DatabaseError(e.into()))?
```
Aeledfyr, line 98: Formatting

Aeledfyr (changes requested):
Putting comments on the formatting issues I found

Scotsguy (approved):

Aeledfyr (approved):
Looks good to me

PR #81, opened 2020-10-18
Add S3 File Host

Geometrically:


src/file_hosting/s3_host.rs
```diff
@@ -0,0 +1,84 @@
+use crate::file_hosting::{DeleteFileData, FileHost, FileHostingError, UploadFileData};
+use async_trait::async_trait;
+use s3::bucket::Bucket;
+
+pub struct S3Host {
+    pub bucket: Bucket,
+}
```
Aeledfyr, line 7: This should have a `new` function that takes a `Bucket` instead of having a `pub` field and using the struct constructor.  Using a constructor allows us to add new fields to this without breaking everywhere we create it.

Aeledfyr (changes requested):
Don't forget to run `cargo fmt`.  I'm not sure exactly why the test is failing, but it looks like it isn't getting the correct environment variables: It's attempting to connect to `https://.***/test.txt` and failing.

src/file_hosting/mod.rs
```diff
@@ -3,16 +3,21 @@ use thiserror::Error;
...
+use s3::S3Error;
 
 #[derive(Error, Debug)]
 pub enum FileHostingError {
     #[error("Error while accessing the data from backblaze")]
     HttpError(#[from] reqwest::Error),
     #[error("Backblaze error: {0}")]
     BackblazeError(serde_json::Value),
+    #[error("S3 error")]
+    S3Error(#[from] S3Error),
```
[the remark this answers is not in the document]
Geometrically: This doesn't work

Aeledfyr (changes requested):
Just some changes for the handling of the environment variables

src/main.rs
```diff
@@ -255,22 +255,24 @@ fn check_env_vars() {
...
 
-    let storage_backend = &*dotenv::var("STORAGE_BACKEND")
-        .ok()
-        .unwrap_or("local".to_string());
+    check_var::<String>("STORAGE_BACKEND");
 
-    if storage_backend == "backblaze" {
+    let storage_backend = dotenv::var("STORAGE_BACKEND").ok();
+
+    if storage_backend == Some("backblaze".to_string()) {
```
Aeledfyr, line 262: These should be `storage_backend.as_ref() == Some("backblaze")` to avoid the allocation;  Same for the other 2 comparisons

Aeledfyr (approved):

PR #83, opened 2020-10-18
Support for using a master key

Geometrically:
This adds the ability to use a Master Key for meilisearch.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -115,11 +115,11 @@ async fn main() -> std::io::Result<()> {
     );
 
     let pool_ref = pool.clone();
-    let search_config_ref = search_config.clone();
+    let thread_search_config = search_config.clone();
     let mut skip = skip_initial;
     scheduler.run(local_index_interval, move || {
         let pool_ref = pool_ref.clone();
-        let search_config_ref = search_config_ref.clone();
+        let thread_search_config = thread_search_config.clone();
         let local_skip = skip;
         if skip {
             skip = false;
@@ -133,7 +133,7 @@ async fn main() -> std::io::Result<()> {
                 index_local: true,
                 index_external: false,
             };
-            let result = index_mods(pool_ref, settings, &search_config_ref).await;
+            let result = index_mods(pool_ref, settings, &thread_search_config).await;
             if let Err(e) = result {
                 warn!("Local mod indexing failed: {:?}", e);
             }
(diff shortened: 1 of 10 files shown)


src/search/mod.rs
```diff
@@ -135,7 +135,8 @@ impl Document for ResultSearchMod {
 
 pub async fn search_for_mod(info: &SearchRequest) -> Result<SearchResults, SearchError> {
     let address = &*dotenv::var("MEILISEARCH_ADDR")?;
-    let client = Client::new(address, "");
+    let key = &*dotenv::var("MEILISEARCH_KEY")?;
+    let client = Client::new(address, key);
```
Aeledfyr, line 139: This may be a good time to move the address and key into a separate structure so that they don't have to be looked up on every request.  We should be able to put it in a actix_web `Data` structure that's cached for each thread.  We could also just cache the meilisearch `Client`, but that may be a bit more complicated.

src/routes/mod.rs
```diff
@@ -68,6 +70,7 @@ impl actix_web::ResponseError for ApiError {
             ApiError::DatabaseError(..) => actix_web::http::StatusCode::INTERNAL_SERVER_ERROR,
             ApiError::AuthenticationError => actix_web::http::StatusCode::UNAUTHORIZED,
             ApiError::JsonError(..) => actix_web::http::StatusCode::BAD_REQUEST,
+            ApiError::SearchError(..) => actix_web::http::StatusCode::FAILED_DEPENDENCY,
```
Aeledfyr, line 73: This error code doesn't appear to be accurate for this error.  `INTERAL_SERVER_ERROR` seems more correct, or something like 503 `SERVICE_UNAVAILABLE` if meilisearch is down.

src/main.rs
```diff
@@ -106,9 +115,11 @@ async fn main() -> std::io::Result<()> {
     );
 
     let pool_ref = pool.clone();
+    let search_config_ref = search_config.clone();
```
Aeledfyr, line 118: This seems annoyingly inefficient, but I'm not sure if there's any way of improving this.  It's only cloning it once per time it runs, so it shouldn't be too bad.  My only comment is that the naming is inaccurate:  the `_ref` implies that this is some sort of shared reference, such as an `Arc`, while this is just a cloned structure.  Something like `thread_search_config` may be more accurate.

Aeledfyr (changes requested):
Looks mostly good besides a few style requests.  On the status codes, it's generally a good idea to refer to [a reference](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status) whenever you have to choose one (besides 200/404/500).

Aeledfyr (approved):

PR #87, opened 2020-10-20
Fix another bug

Geometrically:


Aeledfyr (approved):
Since this occurs in many places we may want to make this a method on the config struct, but it's probably better to merge this first.

PR #89, opened 2020-10-21
Add way to fetch team members

Geometrically:
This also fixes a bug where VersionFiles were not being returned in the route

### DIFF
--- a/src/database/models/team_item.rs
+++ b/src/database/models/team_item.rs
@@ -75,7 +75,13 @@ pub struct TeamMember {
 }
 
 impl TeamMember {
-    pub async fn get_from_team<'a, 'b, E>(id: TeamId,  executor: E) -> Result<Vec<TeamMember>, super::DatabaseError> where E: sqlx::Executor<'a, Database = sqlx::Postgres>,{
+    pub async fn get_from_team<'a, 'b, E>(
+        id: TeamId,
+        executor: E,
+    ) -> Result<Vec<TeamMember>, super::DatabaseError>
+    where
+        E: sqlx::Executor<'a, Database = sqlx::Postgres>,
+    {
         use futures::stream::TryStreamExt;
 
         let team_members = sqlx::query!(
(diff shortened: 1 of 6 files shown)


src/database/models/team_item.rs
```diff
@@ -73,3 +73,31 @@ pub struct TeamMember {
...
+
+        let team_members = sqlx::query!(
+            "
+            SELECT id, user_id, member_name, role
+            FROM team_members
+            WHERE team_id = $1
+            ",
+            id as TeamId,
+        ).fetch_many(executor)
+            .try_filter_map(|e| async {
```
Aeledfyr, line 89: This should be `StreamExt::filter_map` rather than `TryStreamExt::try_filter_map` and `Ok` wrapping the result.

src/routes/teams.rs
```diff
@@ -0,0 +1,27 @@
...
+
+#[get("{id}/members")]
+pub async fn team_members_get(
+    info: web::Path<(TeamId,)>,
+    pool: web::Data<PgPool>,
+) -> Result<HttpResponse, ApiError> {
+    let id = info.into_inner().0;
+    let members_data = TeamMember::get_from_team(id.into(), &**pool)
+        .await
+        .map_err(|e| ApiError::DatabaseError(e.into()))?;
```
Aeledfyr, line 15: Since `get_from_team` returns `database::models::DatabaseError`, this could just be replaced with `.await?;`.  `ApiError` implements `From<DatabaseError>`;  the other places need this because those database methods return `sqlx::Error`.

src/routes/teams.rs
```diff
@@ -0,0 +1,27 @@
...
+        .into_iter()
+        .map(|data| crate::models::teams::TeamMember {
+            user_id: data.user_id.into(),
+            name: data.name,
+            role: data.role
+        })
+        .collect();
+
+    Ok(HttpResponse::Ok().json(team_members))
+}
```
Aeledfyr, line 27: Missing newline @ the end; make sure to run `cargo fmt`.

Aeledfyr (changes requested):

src/routes/teams.rs
Geometrically: Doesn't work, sadly

src/database/models/team_item.rs
Geometrically: The rest of the database uses the Try version, why not use it here?

Aeledfyr: It doesn't accurately represent what the method is doing?  It's like using `Result<T, ()>` to represent an optional value rather than just using `Option<T>`.  `TryStreamExt` is an *extension* of `StreamExt`, so using the original and less complicated version is a good idea.

Aeledfyr (approved):

PR #90, opened 2020-10-25
Rate limiting + version fixes

Geometrically:


src/routes/version_creation.rs
```diff
@@ -70,8 +70,8 @@ pub fn check_version(version: &InitialVersionData) -> Result<(), CreateError> {
     Ok(())
 }
 
-// under `/api/v1/mod/{mod_id}`
-#[post("version")]
+// under `/api/v1/version`
+#[post("")]
```
Aeledfyr, line 74: This may not work the way you expect it to.  In every other place that we do this, the route is specified as `#[post("version")]` and is outside of the `version` scope: https://github.com/modrinth/labrinth/blob/f3c61870dcc80620ade00193e44b7d179742b2e1/src/routes/mod.rs#L21-L30

Cargo.toml
```diff
@@ -50,3 +50,9 @@ git = "https://github.com/launchbadge/sqlx/"
 branch = "master"
 default-features = false
 features = ["runtime-actix", "postgres", "chrono", "offline"]
+
```
Aeledfyr, line 53: Formatting;  This should probably be a single newline, and there should be a newline at the end of the file for consistency.

Cargo.toml
```diff
@@ -50,3 +50,9 @@ git = "https://github.com/launchbadge/sqlx/"
...
+
+[dependencies.actix-ratelimit]
+# Temp fork until actix-ratelimit bumps version on cargo
+git = "https://github.com/TerminalWitchcraft/actix-ratelimit"
+branch = "master"
```
Aeledfyr, line 58: 1. This isn't a fork of `actix-ratelimit`, that's the official repo
2.  This should specify the commit hash using `rev` instead of setting the branch to `master` so that it doesn't break in the future.
    - In this case it would be `rev = "870822067dfeae7cc0304352d81c4cb79ee27f5a"`

src/database/models/version_item.rs
```diff
@@ -84,6 +84,17 @@ impl VersionBuilder {
...
+        sqlx::query!(
+                "
+                UPDATE mods
+                SET updated = NOW()
+                WHERE id = $1
+                ",
+                self.mod_id as ModId,
+            )
+            .execute(&mut *transaction)
+            .await?;
```
Aeledfyr, line 96: The indentation is messed up here; fix it and run `cargo fmt` to get the rest of the formatting issues.

Aeledfyr (changes requested):
A few formatting changes, but the rest looks good

Aeledfyr (approved):
Looks good.  I can't verify the correctness of the rate limiter, but that should be testable.

PR #91, opened 2020-10-27
Game Version types, indexing, and bugfixes

Aeledfyr:
This adds type filtering to the game versions routes so that it should be possible to only show release versions.  Other changes:

- All tag create/edit routes now are (should be) idempotent
- Fix version numbers for mods added to the index through the recently created mod queue
- Automatically check the launchermeta every 6 hours for new minecraft versions
- Version type filtering: `/api/v1/categories/game_versions?type=release` should only give versions tagged as a release by Mojang

### DIFF
--- a/migrations/20201021214908_extend-game-version.sql
+++ b/migrations/20201021214908_extend-game-version.sql
@@ -1,0 +1,3 @@
+
+ALTER TABLE game_versions
+ADD COLUMN type varchar(16) NOT NULL DEFAULT 'other';
--- a/sqlx-data.json
+++ b/sqlx-data.json
@@ -278,26 +278,6 @@
       ]
     }
   },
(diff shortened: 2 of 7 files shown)


Geometrically (approved):
Looks good! However does it handle the old versions that I put manually into the DB? Or do I have to drop all the columns myself

Aeledfyr:
The current versions in the database will be kept (the `type` column defaults to "other") and when the version indexing runs it will update the existing versions to have the proper type.  The only problematic ones are the ones with spaces in their names;  the indexer doesn't add these so they have to be processed manually.  Mainly I have to remove the 1.14 prereleases and add the properly formatted versions of them, and find a reasonable name for the shareware april fools snapshot.  (My account needs to be an admin to be able to fix that, though).
Creating versions through the API still works, but you have to send a JSON body with `{"type":"release"}`.

PR #92, opened 2020-10-28
Team routes

Geometrically:


Redblueflame:
Resolves #94

src/routes/mod_creation.rs
```diff
@@ -434,6 +434,8 @@ async fn mod_create_inner(
                 user_id: current_user.id.into(),
                 name: current_user.username.clone(),
                 role: crate::models::teams::OWNER_ROLE.to_owned(),
+                permissions: 0x1 | 0x2,
```
Aeledfyr, line 437: For this to be usable, this must either use something like [bitflags](https://docs.rs/bitflags) or have a set of constants;  This gives no indication what 0x1 and 0x2 mean.

src/routes/teams.rs
```diff
@@ -18,8 +21,136 @@ pub async fn team_members_get(
             user_id: data.user_id.into(),
             name: data.name,
             role: data.role,
+            permissions: data.permissions as u64
```
Aeledfyr, line 24: This doesn't seem like it should be public, besides from what user is the owner

src/routes/teams.rs
```diff
@@ -18,8 +21,136 @@ pub async fn team_members_get(
...
+#[post("{id}/join")]
+pub async fn join_team(
+    req: HttpRequest,
+    info: web::Path<(TeamId,)>,
+    pool: web::Data<PgPool>,
+) -> Result<HttpResponse, ApiError> {
+    let team_id = info.into_inner().0.into();
+    let current_user = get_user_from_headers(req.headers(), &**pool).await.map_err(|_| ApiError::AuthenticationError)?;
+
+    TeamMember::edit_team_member(team_id, current_user.id.into(), None, None, Some(true), &**pool).await?;
```
Aeledfyr, line 40: This should use something like a `TeamMemberBuilder`, since the function arguments give no indication of what the arguments actually mean.  Also, shouldn't this verify that there's a pending request first?

src/routes/teams.rs
```diff
@@ -18,8 +21,136 @@ pub async fn team_members_get(
...
+) -> Result<HttpResponse, ApiError> {
+    let team_id = info.into_inner().0.into();
+
+    let mut transaction = pool.begin().await.map_err(|e| ApiError::DatabaseError(e.into()))?;
+
+    let current_user = get_user_from_headers(req.headers(), &**pool).await.map_err(|_| ApiError::AuthenticationError)?;
+    let team_member = TeamMember::get_from_user_id(team_id, current_user.id.into(), &**pool).await?;
+
+    if let Some(member) = team_member {
+        if (member.permissions & (1 << 4)) != 0 && new_member.role != crate::models::teams::OWNER_ROLE {
```
Aeledfyr, line 60: There shouldn't be a special case for the "owner" role, besides ensuring that the owner always has all permissions.  The permissions bit here should be defined as a constant and be documented, or use bitflags.

src/routes/teams.rs
```diff
@@ -18,8 +21,136 @@ pub async fn team_members_get(
...
+            // TODO: Prevent user from giving another user permissions they do not have
```
Aeledfyr, line 61: This should not be merged until this is implemented.

src/routes/teams.rs
```diff
@@ -18,8 +21,136 @@ pub async fn team_members_get(
...
+) -> Result<HttpResponse, ApiError> {
+    let ids = info.into_inner();
+    let id = ids.0.into();
+    let user_id = ids.1.into();
+
+    let current_user = get_user_from_headers(req.headers(), &**pool).await.map_err(|_| ApiError::AuthenticationError)?;
+    let team_member = TeamMember::get_from_user_id(id, current_user.id.into(), &**pool).await?;
+
+    if let Some(member) = team_member {
+        if (member.permissions & (1 << 6)) != 0 && edit_member.role.clone().unwrap_or("".to_string()) != "Owner".to_string() {
```
Aeledfyr, line 105: This shouldn't have a special case for the owner (and if it has to, that's a very bad way of doing it... just do `edit_member.role != crate::models::teams::OWNER_ROLE`)
Again, permissions bitset needs to be documented.

src/routes/teams.rs
```diff
@@ -18,8 +21,136 @@ pub async fn team_members_get(
...
+            // TODO: Prevent user from giving another user permissions they do not have
```
Aeledfyr, line 106: The same as the above, this shouldn't be merged until this is implemented.  This should also check the role, since if "owner" is still special cased, the new role could be "owner", so multiple team members would be marked as the owner.
Should we have a special system for transferring ownership of a team?

src/routes/users.rs
```diff
@@ -121,6 +121,38 @@ pub async fn mods_list(
...
+
+    let current_user = get_user_from_headers(req.headers(), &**pool).await.map_err(|_| ApiError::AuthenticationError)?;
+
+    let mut results = Vec::new();
+
+    let id_int = id.0 as u64;
+    if current_user.id.0 == id_int {
+        results = crate::database::models::TeamMember::get_from_user_private(id, &**pool).await?;
+    } else {
+        results = crate::database::models::TeamMember::get_from_user_private(id, &**pool).await?;
```
Aeledfyr, line 140: Shouldn't this return an error?

src/routes/users.rs
```diff
@@ -121,6 +121,38 @@ pub async fn mods_list(
     }
 }
 
+#[get("invites")]
+pub async fn remove_team_member(
```
Aeledfyr, line 125: This should be named `list_invites` or something similar

src/routes/users.rs
```diff
@@ -121,6 +121,38 @@ pub async fn mods_list(
...
+) -> Result<HttpResponse, ApiError> {
+    let id: crate::database::models::UserId = info.into_inner().0.into();
+
+    let current_user = get_user_from_headers(req.headers(), &**pool).await.map_err(|_| ApiError::AuthenticationError)?;
+
+    let mut results = Vec::new();
+
+    let id_int = id.0 as u64;
+    if current_user.id.0 == id_int {
+        results = crate::database::models::TeamMember::get_from_user_private(id, &**pool).await?;
```
Aeledfyr, line 138: This returns a list of all team members of a user, both accepted and not accepted.  If this is meant to list unaccepted invites, this should skip the accepted ones.

src/routes/users.rs
```diff
@@ -121,6 +121,38 @@ pub async fn mods_list(
...
+    pool: web::Data<PgPool>,
+) -> Result<HttpResponse, ApiError> {
+    let id: crate::database::models::UserId = info.into_inner().0.into();
+
+    let current_user = get_user_from_headers(req.headers(), &**pool).await.map_err(|_| ApiError::AuthenticationError)?;
+
+    let mut results = Vec::new();
+
+    let id_int = id.0 as u64;
+    if current_user.id.0 == id_int {
```
Aeledfyr, line 137: This comparison could be written as `current_user.id.0 == id.0 as u64`

src/routes/versions.rs
```diff
@@ -156,7 +156,7 @@ pub async fn version_get(
 #[delete("{version_id}")]
 pub async fn version_delete(
     req: HttpRequest,
-    info: web::Path<(models::ids::ModId, models::ids::VersionId)>,
+    info: web::Path<(models::ids::VersionId,)>,
     pool: web::Data<PgPool>,
 ) -> Result<HttpResponse, ApiError> {
     check_is_moderator_from_headers(
```
Aeledfyr, line 162: This should allow users to delete their own mod versions if they have the permissions

src/routes/versions.rs
```diff
@@ -170,7 +170,7 @@ pub async fn version_delete(
     .map_err(|_| ApiError::AuthenticationError)?;
 
     // TODO: check if the mod exists and matches the version id
```
Aeledfyr, line 172: This TODO is no longer applicable, I think

Aeledfyr (changes requested):

src/routes/teams.rs
Geometrically: Why not?

src/routes/teams.rs
Geometrically: What do you mean by pending request? The argument names are selfexplanatory

src/routes/teams.rs
Geometrically: The team member's role can never equal "Owner", besides the mod owner. This is preventing the user from setting "Owner" as their role when they are not the owner

src/routes/teams.rs
Geometrically: A lowercase owner is fine, for consistency I would like any owner of a mod for their role to be "Owner". We can add an ownership transfer soon

src/routes/teams.rs
Aeledfyr: There is no reason that anyone outside of the team should need or have access to a the specific permissions within the team.  That would be a security issue, since it would give people targets for which accounts to take over to get access to a mod, and it would give information about the structure of a team that shouldn't be publicly available.

src/routes/teams.rs
Aeledfyr: `None`, `None`, and `Some(true)` are not self-explanatory.

A pending request means a `TeamMember` with `accepted` being false;  You are currently assuming that there's a currently non accepted team member request.  You should check whether there is a current non accepted team member, and if not return a useful error message.

src/routes/mods.rs
```diff
@@ -130,17 +132,32 @@ pub async fn mod_delete(
...
+            mod_item.team_id,
+            user.id.into(),
+            &**pool,
+        )
+        .await
+        .map_err(|e| ApiError::DatabaseError(e.into()))?
+        .ok_or_else(|| ApiError::InvalidInputError("Invalid Mod ID specified!".to_string()))?;
+
+        if Permissions::from_bits_truncate(team_member.permissions as u64) & Permissions::DELETE_MOD
+            != Permissions::DELETE_MOD
```
Aeledfyr, line 155: This should use `from_bits` rather than `from_bits_truncate` and handle the error.  This should also use `.contains(Permissions::DELETE_MOD)`, as it is more concise.

src/routes/teams.rs
```diff
@@ -1,25 +1,233 @@
...
+        .await
+        .map_err(|_| ApiError::AuthenticationError)?;
+    let team_member =
+        TeamMember::get_from_user_id(team_id, current_user.id.into(), &**pool).await?;
+
+    if let Some(member) = team_member {
+        let permissions = Permissions::from_bits_truncate(member.permissions as u64);
+
+        if permissions & Permissions::MANAGE_INVITES == Permissions::MANAGE_INVITES
+            && new_member.role != crate::models::teams::OWNER_ROLE
```
Aeledfyr, line 103: This could be `permissions.contains(Permissions::MANAGE_INVITES)`

src/routes/teams.rs
```diff
@@ -1,25 +1,233 @@
...
+
+            let new_id =
+                crate::database::models::ids::generate_team_member_id(&mut transaction).await?;
+            TeamMember {
+                id: new_id,
+                team_id,
+                user_id: new_member.user_id.into(),
+                name: new_member.name.clone(),
+                role: new_member.role.clone(),
+                permissions: new_member.permissions.bits() as i64,
```
Aeledfyr, line 117: This needs to be checked

src/routes/teams.rs
```diff
@@ -1,25 +1,233 @@
...
+                } else {
+                    Err(ApiError::AuthenticationError)
+                }
+            } else {
+                if (member.permissions & (1 << 4)) != 0 {
+                    TeamMember::delete(id, user_id, &**pool).await?;
+                    Ok(HttpResponse::Ok().body(""))
+                } else {
+                    Err(ApiError::AuthenticationError)
+                }
```
Aeledfyr, line 225: This should use bitflags methods

src/routes/users.rs
```diff
@@ -121,6 +122,43 @@ pub async fn mods_list(
...
+        results = crate::database::models::TeamMember::get_from_user_public(id, &**pool).await?;
+    }
+
+    let team_members: Vec<crate::models::teams::TeamMember> = results
+        .into_iter()
+        .map(|data| crate::models::teams::TeamMember {
+            user_id: data.user_id.into(),
+            name: data.name,
+            role: data.role,
+            permissions: Permissions::from_bits_truncate(data.permissions as u64),
```
Aeledfyr, line 155: This shouldn't return permissions unless the user is the current logged in user.

src/routes/versions.rs
```diff
@@ -156,21 +158,44 @@ pub async fn version_get(
...
+            user.id.into(),
+            &**pool,
+        )
+        .await
+        .map_err(|e| ApiError::DatabaseError(e.into()))?
+        .ok_or_else(|| ApiError::InvalidInputError("Invalid Version ID specified!".to_string()))?;
+
+        if Permissions::from_bits_truncate(team_member.permissions as u64)
+            & Permissions::DELETE_VERSION
+            != Permissions::DELETE_VERSION
```
Aeledfyr, line 193: This could use `.contains`

src/database/models/team_item.rs
```diff
@@ -100,11 +107,221 @@ impl TeamMember {
...
+        new_accepted: Option<bool>,
+        executor: E,
+    ) -> Result<(), super::DatabaseError>
+    where
+        E: sqlx::Executor<'a, Database = sqlx::Postgres>,
+    {
+        let mut query = "UPDATE team_members".to_string();
+        let mut current_index: i16 = 3;
+
+        if new_permissions.is_some() {
```
Aeledfyr, line 289: What happens if all three are `None`?

src/database/models/team_item.rs
```diff
@@ -72,6 +77,8 @@ pub struct TeamMember {
     /// The name of the user
     pub name: String,
     pub role: String,
+    pub permissions: i64,
```
Aeledfyr, line 80: This should be the bitflags type wrapper

Aeledfyr (changes requested):

src/routes/teams.rs
```diff
@@ -1,25 +1,233 @@
...
+        .await
+        .map_err(|_| ApiError::AuthenticationError)?;
+    let team_member = TeamMember::get_from_user_id(id, current_user.id.into(), &**pool).await?;
+
+    if let Some(member) = team_member {
+        let permissions = Permissions::from_bits_truncate(member.permissions as u64);
+
+        if permissions & Permissions::EDIT_MEMBER == Permissions::EDIT_MEMBER
+            && edit_member.role.clone().unwrap_or("".to_string())
+                != crate::models::teams::OWNER_ROLE.to_string()
```
Aeledfyr, line 165: This should be `edit_member.role.as_deref() != Some(crate::models::teams::OWNER_ROLE)` or something similar;  This shouldn't have `to_string` or `clone`.

src/routes/teams.rs
Aeledfyr: If this is fixed, this TODO should be removed

src/database/models/team_item.rs
Geometrically: Mod isn't edited, nothing happens, intended.

Aeledfyr: Is this valid SQL?  I think this would cause an error.  This should check if all three are empty, and in that case just return.
```sql
UPDATE team_members
WHERE ...
```

Redblueflame: And in that case, we have an issue

src/routes/mods.rs
Geometrically: I switched to contains. Why from bits? There is no need for an error, as we can still get the permissions out of the bits. Some languages may use the on bit as default, so we shouldn't throw an error for that

Aeledfyr: This should be `from_bits` because it should fail if there is anything incorrect with the permissions bitset.  For example, if we add new permissions and accidentally run on an older version of the backend with newer permissions in the database, we want to be careful to handle that well.  In this case it should be fine to use `from_bits_truncate` as it's only checking permissions, but everywhere else should use `from_bits` and check for errors.  If we don't check for errors, we may lose data, as the bitset will silently truncate the newer permissions.

Geometrically: Isn't this line checking for permissions though, for the editing and create I can change to from bits.

src/database/models/team_item.rs
```diff
@@ -36,18 +39,21 @@ impl TeamBuilder {
...
+                permissions: member.permissions,
+                accepted: member.accepted,
             };
 
             sqlx::query!(
                 "
-                INSERT INTO team_members (id, team_id, user_id, member_name, role)
-                VALUES ($1, $2, $3, $4, $5)
+                INSERT INTO team_members (id, team_id, user_id, member_name, role, permissions, accepted)
+                VALUES ($1, $2, $3, $4, $5, $6, TRUE)
```
Aeledfyr, line 49: This should not be hardcoded to be `TRUE`.  Each team member has an individual accepted field, and we should be able to create teams with an initial owner (`accepted = true`) as well as initial team member requests (`accepted = false`).

src/database/models/team_item.rs
```diff
@@ -100,11 +108,230 @@ impl TeamMember {
                 user_id: UserId(m.user_id),
                 name: m.member_name,
                 role: m.role,
+                permissions: Permissions::from_bits_truncate(m.permissions as u64),
```
Aeledfyr, line 111: This should not truncate, since it's loading the data; using truncate when checking permissions is fine, but using it here could cause data loss.

src/database/models/team_item.rs
```diff
@@ -100,11 +108,230 @@ impl TeamMember {
...
+        )
+        .fetch_many(executor)
+        .try_filter_map(|e| async {
+            Ok(e.right().map(|m| TeamMember {
+                id: TeamMemberId(m.id),
+                team_id: TeamId(m.team_id),
+                user_id: id,
+                name: m.member_name,
+                role: m.role,
+                permissions: Permissions::from_bits_truncate(m.permissions as u64),
```
Aeledfyr, line 146: Same as above, this shouldn't truncate.

src/database/models/team_item.rs
```diff
@@ -100,11 +108,230 @@ impl TeamMember {
...
```
Aeledfyr, line 181: Same as above, this shouldn't truncate.

src/database/models/team_item.rs
```diff
@@ -100,11 +108,230 @@ impl TeamMember {
...
+        .await?;
+
+        if let Some(m) = result {
+            Ok(Some(TeamMember {
+                id: TeamMemberId(m.id),
+                team_id: id,
+                user_id,
+                name: m.member_name,
+                role: m.role,
+                permissions: Permissions::from_bits_truncate(m.permissions as u64),
```
Aeledfyr, line 218: Same as above, this shouldn't truncate.

src/database/models/team_item.rs
```diff
@@ -100,11 +108,230 @@ impl TeamMember {
...
+            "
+            DELETE FROM team_members
+            WHERE (team_id = $1 AND user_id = $2 AND NOT role = $3)
+            ",
+            id as TeamId,
+            user_id as UserId,
+            crate::models::teams::OWNER_ROLE
+        )
+        .execute(executor)
+        .await?;
```
Aeledfyr, line 271: This should return an error if nothing was deleted (ie. attempting to delete the owner).  I'm not sure exactly how to go about that, but it seems like you should be able to get the number of affected rows; if it's anything other than one, return an error.

Aeledfyr: We may also want a separate method for rejecting requests that does the same as this method but requires that `accepted` be false.  While we could use this method, having a separate method/query could help prevent accidental race conditions with accepting / rejecting a team member request.

src/database/models/team_item.rs
```diff
@@ -100,11 +108,230 @@ impl TeamMember {
...
+            sqlx::query!(
+                "
+                UPDATE team_members
+                SET permissions = $1
+                WHERE (team_id = $2 AND user_id = $3 AND NOT role = $4)
+                ",
+                permissions,
+                id as TeamId,
+                user_id as UserId,
+                crate::models::teams::OWNER_ROLE.to_string(),
```
Aeledfyr, line 297: `.to_string()` shouldn't be needed here, as it wasn't used above.

src/database/models/team_item.rs
```diff
@@ -100,11 +108,230 @@ impl TeamMember {
...
+            sqlx::query!(
+                "
+                UPDATE team_members
+                SET role = $1
+                WHERE (team_id = $2 AND user_id = $3 AND NOT role = $4)
+                ",
+                role,
+                id as TeamId,
+                user_id as UserId,
+                crate::models::teams::OWNER_ROLE.to_string(),
```
Aeledfyr, line 313: `.to_string()` shouldn't be needed here, as it wasn't used above.

src/database/models/team_item.rs
```diff
@@ -100,11 +108,230 @@ impl TeamMember {
...
+            sqlx::query!(
+                "
+                UPDATE team_members
+                SET accepted = $1
+                WHERE (team_id = $2 AND user_id = $3 AND NOT role = $4)
+                ",
+                accepted,
+                id as TeamId,
+                user_id as UserId,
+                crate::models::teams::OWNER_ROLE.to_string(),
```
Aeledfyr, line 329: `.to_string()` shouldn't be needed here, as it wasn't used above.

src/database/models/team_item.rs
```diff
@@ -100,11 +108,230 @@ impl TeamMember {
...
+        .execute(executor)
+        .await?;
+
+        Ok(())
+    }
+
+    pub async fn edit_team_member<'a, 'b, E>(
+        id: TeamId,
+        user_id: UserId,
+        new_permissions: Option<i64>,
```
Aeledfyr, line 279: This should take the permissions bitset rather than just an integer.  Using an integer here allows us to potentially insert invalid bit patterns into the database.

src/database/models/team_item.rs
```diff
@@ -100,11 +108,230 @@ impl TeamMember {
...
+                role,
+                id as TeamId,
+                user_id as UserId,
+                crate::models::teams::OWNER_ROLE.to_string(),
+            )
+            .execute(executor)
+            .await?;
+        }
+
+        if let Some(accepted) = new_accepted {
```
Aeledfyr, line 319: This should only ever work for accepting a request.  You should not be able to change accepted to false; leaving a team should be deleting the team member.

src/models/teams.rs
```diff
@@ -20,6 +20,27 @@ pub struct Team {
     pub members: Vec<TeamMember>,
 }
 
+bitflags::bitflags! {
+    #[derive(Serialize, Deserialize)]
+    pub struct Permissions: u64 {
```
Aeledfyr, line 25: It's not a requirement, but it could be useful to use `1 << n` notation for this as we add more permissions flags.

src/models/teams.rs
```diff
@@ -29,4 +50,6 @@ pub struct TeamMember {
     pub name: String,
     /// The role of the user in the team
     pub role: String,
+    /// A bitflag containing the user's permissions in this team
```
Aeledfyr, line 53: This should say `A bitset`

src/routes/mods.rs
```diff
@@ -130,17 +132,30 @@ pub async fn mod_delete(
...
+        let team_member = database::models::TeamMember::get_from_user_id(
+            mod_item.team_id,
+            user.id.into(),
+            &**pool,
+        )
+        .await
+        .map_err(|e| ApiError::DatabaseError(e.into()))?
+        .ok_or_else(|| ApiError::InvalidInputError("Invalid Mod ID specified!".to_string()))?;
+
+        if team_member.permissions.contains(Permissions::DELETE_MOD) {
```
Aeledfyr, line 154: This condition should be inverted.

src/routes/teams.rs
```diff
@@ -1,25 +1,227 @@
...
             role: data.role,
+            permissions: Permissions::default(),
         })
         .collect();
 
     Ok(HttpResponse::Ok().json(team_members))
 }
+
+#[post("{id}/join")]
+pub async fn join_team(
```
Aeledfyr, line 55: This needs to check for the existence of an invite and return useful error messages for the cases where: there is no pending invite, you are already a member of that team, or the team doesn't exist

src/routes/teams.rs
```diff
@@ -1,25 +1,227 @@
...
+        .map_err(|_| ApiError::AuthenticationError)?;
+    let team_member =
+        TeamMember::get_from_user_id(team_id, current_user.id.into(), &**pool).await?;
+
+    if let Some(member) = team_member {
+        if member.permissions.contains(Permissions::MANAGE_INVITES)
+            && new_member.role != crate::models::teams::OWNER_ROLE
+        {
+            if !member.permissions.contains(new_member.permissions) {
+                return Err(ApiError::AuthenticationError);
```
Aeledfyr, line 104: This should return a more specific error.

src/routes/teams.rs
```diff
@@ -1,25 +1,227 @@
...
+            .map_err(|e| ApiError::DatabaseError(e.into()))?;
+
+            transaction
+                .commit()
+                .await
+                .map_err(|e| ApiError::DatabaseError(e.into()))?;
+
+            Ok(HttpResponse::Ok().body(""))
+        } else {
+            Err(ApiError::AuthenticationError)
```
Aeledfyr, line 129: This should return a more specific error.

src/routes/teams.rs
```diff
@@ -1,25 +1,227 @@
...
+                return Err(ApiError::AuthenticationError);
+            }
+
+            let new_id =
+                crate::database::models::ids::generate_team_member_id(&mut transaction).await?;
+            TeamMember {
+                id: new_id,
+                team_id,
+                user_id: new_member.user_id.into(),
+                name: new_member.name.clone(),
```
Aeledfyr, line 113: I don't think that it's a good idea to have the inviting user be able to specify how the new member's username will appear... This should just use the other member's username as a default, and allow them to change it later.

src/routes/teams.rs
```diff
@@ -1,25 +1,227 @@
...
+
+    Ok(HttpResponse::Ok().body(""))
+}
+
+#[post("{id}/members")]
+pub async fn add_team_member(
+    req: HttpRequest,
+    info: web::Path<(TeamId,)>,
+    pool: web::Data<PgPool>,
+    new_member: web::Json<crate::models::teams::TeamMember>,
```
Aeledfyr, line 84: This should have a separate struct for creating a member, with most fields being optional.  `permissions` and `role` should be optional and have reasonable defaults, and `name` (if you even allow it to be specified) should default to the new user's username.

src/routes/teams.rs
```diff
@@ -1,25 +1,227 @@
...
+        } else {
+            Err(ApiError::AuthenticationError)
+        }
+    } else {
+        Err(ApiError::AuthenticationError)
+    }
+}
+
+#[derive(Serialize, Deserialize, Clone)]
+pub struct EditTeamMember {
```
Aeledfyr, line 137: Editing a team member should also allow modifying the team member's `name`.

src/routes/teams.rs
```diff
@@ -1,25 +1,227 @@
...
+    let user_id = ids.1.into();
+
+    let current_user = get_user_from_headers(req.headers(), &**pool)
+        .await
+        .map_err(|_| ApiError::AuthenticationError)?;
+    let team_member = TeamMember::get_from_user_id(id, current_user.id.into(), &**pool).await?;
+
+    if let Some(member) = team_member {
+        if member.permissions.contains(Permissions::EDIT_MEMBER)
+            && edit_member.role.as_deref() != Some(crate::models::teams::OWNER_ROLE)
```
Aeledfyr, line 160: There needs to be some way for a user to modify their own name as a team member; it seems like it could fit in this method, but you would have to be careful to make sure that the permissions and role are not changed.

src/routes/teams.rs
```diff
@@ -1,25 +1,227 @@
...
+            if let Some(new_permissions) = edit_member.permissions {
+                if !member.permissions.contains(new_permissions) {
+                    return Err(ApiError::AuthenticationError);
+                }
+            }
+
+            TeamMember::edit_team_member(
+                id,
+                user_id,
+                edit_member.permissions.map(|x| x.bits() as i64),
```
Aeledfyr, line 171: This conversion should happen within the `edit_team_member` method rather than outside of it to prevent any accidental errors.

src/routes/teams.rs
```diff
@@ -1,25 +1,227 @@
...
+    let id = ids.0.into();
+    let user_id = ids.1.into();
+
+    let current_user = get_user_from_headers(req.headers(), &**pool)
+        .await
+        .map_err(|_| ApiError::AuthenticationError)?;
+    let team_member = TeamMember::get_from_user_id(id, current_user.id.into(), &**pool).await?;
+
+    if let Some(member) = team_member {
+        let delete_member_option = TeamMember::get_from_user_id(id, user_id, &**pool).await?;
```
Aeledfyr, line 203: This looks the exact same as the query for the team member above.  Are we ever actually checking that members have `accepted = true` before they can use the permissions for that team member?  The `permissions` bitset for a non accepted team member should be treated like it has no permissions.

src/routes/teams.rs
```diff
@@ -1,25 +1,227 @@
...
+
+        if let Some(delete_member) = delete_member_option {
+            if delete_member.accepted {
+                if member.permissions.contains(Permissions::DELETE_MOD) {
```
Aeledfyr, line 207: This needs to have different permission levels depending on the scenario.
This should normally be `Permissions::REMOVE_MEMBER`.

If the user doing the action is the same as the user for the team member being removed, this shouldn't require any permissions: users should be able to freely leave teams.

The owner should not be deleted through this route.

src/routes/teams.rs
```diff
@@ -1,25 +1,227 @@
...
+                    TeamMember::delete(id, user_id, &**pool).await?;
+                    Ok(HttpResponse::Ok().body(""))
+                } else {
+                    Err(ApiError::AuthenticationError)
+                }
+            } else {
+                if member.permissions.contains(Permissions::MANAGE_INVITES) {
```
Aeledfyr, line 214: This should allow users to reject team member requests; if the user is the same as the user for the team member being deleted, this shouldn't require any permissions.

src/routes/users.rs
```diff
@@ -121,6 +122,44 @@ pub async fn mods_list(
...
+    }
+
+    let team_members: Vec<crate::models::teams::TeamMember> = results
+        .into_iter()
+        .map(|data| crate::models::teams::TeamMember {
+            user_id: data.user_id.into(),
+            name: data.name,
+            role: data.role,
+            // TODO: Return permissions if user is the logged in user
+            permissions: Permissions::default(),
```
Aeledfyr, line 156: This could be implemented like the following:
```rust
let same_user = current_user.map(|u| u.id.0) == Some(id.0 as u64);
...
        .map(|data| crate::models::teams::TeamMember {
            user_id: data.user_id.into(),
            name: data.name,
            role: data.role,
            permissions: if same_user { data.permissions } else { Permissions::default() },
        }
```
The permissions field in `TeamMember` should be optional so that we can more accurately not specify the permissions rather than specifying  the default permissions.

Aeledfyr (changes requested):
One major potential issue with this system of team member requests is that we don't always check whether `accepted` is `true`.  Non accepted team members should never be able to use those permissions, but I believe that with the current system that isn't being checked.  Team members that haven't been accepted should not be treated like members of that team, so we need to have different `TeamMember::get_from_user_id` methods.  `get_from_user_id` should only return accepted members, and we should add a separate method for getting only pending requests or either pending or accepted members.

src/database/models/team_item.rs
```diff
@@ -100,11 +108,230 @@ impl TeamMember {
...
+        Ok(())
+    }
+
+    pub async fn edit_team_member<'a, 'b, E>(
+        id: TeamId,
+        user_id: UserId,
+        new_permissions: Option<i64>,
+        new_role: Option<String>,
+        new_accepted: Option<bool>,
+        executor: E,
```
Aeledfyr, line 282: This executor parameter may have to change, since this should only be used with transactions.  Look at the mod builder for example parameters.

src/routes/teams.rs
Geometrically: No, the team member's name is their username

src/routes/teams.rs
Geometrically: No, the team member's name is their username

src/routes/teams.rs
Aeledfyr: Then why do we have it separately defined?  And why can the person inviting a team member specify the name that the invited user ends up with?

src/routes/teams.rs
Geometrically: Fixed, the owner isn't able to be deleted.

src/routes/teams.rs
Geometrically: I think it's fine. This is called when the permissions check is failed, so the user is unauthorized

src/routes/teams.rs
Geometrically: No pending invite/Team doesn't exist have the SQLX error messages, which are sufficient, and the other two have no harm as accepted just gets set to true

src/database/models/team_item.rs
```diff
@@ -86,25 +95,287 @@ impl TeamMember {
...
+                Ok(None)
+            }
+        })
+        .try_collect::<Vec<TeamMember>>()
+        .await?;
+
+        Ok(team_members)
+    }
+
+    pub async fn get_from_user_id<'a, 'b, E>(
```
Aeledfyr, line 219: This should document that it only retrieves accepted team members

src/routes/mods.rs
```diff
@@ -130,17 +132,30 @@ pub async fn mod_delete(
...
+        let team_member = database::models::TeamMember::get_from_user_id(
+            mod_item.team_id,
+            user.id.into(),
+            &**pool,
+        )
+        .await
+        .map_err(|e| ApiError::DatabaseError(e.into()))?
+        .ok_or_else(|| ApiError::InvalidInputError("Invalid Mod ID specified!".to_string()))?;
+
+        if !team_member.permissions.contains(Permissions::DELETE_MOD) && team_member.accepted {
```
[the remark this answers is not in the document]
Aeledfyr: Actually, this won't fail since `get_from_user_id` can only return accepted users.  The `team_member.accepted` condition should be removed.

src/routes/teams.rs
Aeledfyr: The sqlx error messages are not sufficient for a proper API.  If you are already a member of that team, it should still return an error.

src/routes/teams.rs
```diff
@@ -1,25 +1,275 @@
...
+
+    let current_user = get_user_from_headers(req.headers(), &**pool)
+        .await
+        .map_err(|_| ApiError::AuthenticationError)?;
+    let team_member =
+        TeamMember::get_from_user_id(team_id, current_user.id.into(), &**pool).await?;
+
+    if let Some(member) = team_member {
+        if member.permissions.contains(Permissions::MANAGE_INVITES)
+            && member.accepted
```
Aeledfyr, line 125: This condition should be removed.  `get_from_user_id` can only return accepted team members.

src/routes/teams.rs
```diff
@@ -1,25 +1,275 @@
...
+                crate::database::models::ids::generate_team_member_id(&mut transaction).await?;
+            TeamMember {
+                id: new_id,
+                team_id,
+                user_id: new_member.user_id.into(),
+                name: new_user.username,
+                role: new_member.role.clone(),
+                permissions: new_member.permissions,
+                accepted: false,
+            }
```
Aeledfyr, line 150: This needs to check if the user is already part of that team and return an error.

src/routes/teams.rs
Aeledfyr: For a proper API, we should give specifics for why the call failed.  In this case, it should be `"You do not have permission to add new members to this team"`.  The check whether the new role is "Owner" should be moved and have its own error message.

src/routes/teams.rs
```diff
@@ -1,25 +1,275 @@
...
+    let team_member = TeamMember::get_from_user_id(id, current_user.id.into(), &**pool).await?;
+
+    let mut transaction = pool
+        .begin()
+        .await
+        .map_err(|e| ApiError::DatabaseError(e.into()))?;
+
+    if let Some(member) = team_member {
+        if member.permissions.contains(Permissions::EDIT_MEMBER)
+            && member.accepted
```
Aeledfyr, line 199: This condition should be removed. `get_from_user_id` can only return accepted team members.

src/routes/teams.rs
```diff
@@ -1,25 +1,275 @@
...
+            .await?;
+
+            transaction
+                .commit()
+                .await
+                .map_err(|e| ApiError::DatabaseError(e.into()))?;
+
+            Ok(HttpResponse::Ok().body(""))
+        } else {
+            Err(ApiError::AuthenticationError)
```
Aeledfyr, line 226: See above, this should return a useful error message.

src/routes/teams.rs
```diff
@@ -1,25 +1,275 @@
...
+        .map_err(|_| ApiError::AuthenticationError)?;
+    let team_member = TeamMember::get_from_user_id(id, current_user.id.into(), &**pool).await?;
+
+    let mut transaction = pool
+        .begin()
+        .await
+        .map_err(|e| ApiError::DatabaseError(e.into()))?;
+
+    if let Some(member) = team_member {
+        if member.permissions.contains(Permissions::EDIT_MEMBER)
```
Aeledfyr, line 198: Users should be able to modify their own names within the team without the `EDIT_MEMBER` permission.

src/routes/teams.rs
Aeledfyr: `get_from_user_id` can only return accepted team members, so this route cannot remove pending team member requests.  We need to add a new function for getting potentially pending team members and use that here;  `get_from_user_id` must stay the same and only return accepted members because it is used for checking permissions.

src/routes/teams.rs
```diff
@@ -1,25 +1,275 @@
...
+
+        if let Some(delete_member) = delete_member_option {
+            if delete_member.accepted {
+                if delete_member.user_id == member.user_id
+                    || (member.permissions.contains(Permissions::REMOVE_MEMBER) && member.accepted)
+                {
+                    TeamMember::delete(id, user_id, &**pool).await?;
+                    Ok(HttpResponse::Ok().body(""))
+                } else {
+                    Err(ApiError::AuthenticationError)
```
Aeledfyr, line 259: All of these `AuthenticationError`s should have error messages describing what went wrong.

src/routes/users.rs
```diff
@@ -121,6 +122,49 @@ pub async fn mods_list(
...
+
+    let current_user = get_user_from_headers(req.headers(), &**pool).await.ok();
+
+    let results;
+    let mut same_user = false;
+
+    if let Some(user) = current_user {
+        if user.id.0 == id.0 as u64 {
+            results =
+                crate::database::models::TeamMember::get_from_user_private(id, &**pool).await?;
```
Aeledfyr, line 141: Since this is used a lot, it may be worth having `use crate::database::models::TeamMember;`

src/routes/versions.rs
```diff
@@ -156,21 +158,45 @@ pub async fn version_get(
...
+            &**pool,
+        )
+        .await
+        .map_err(|e| ApiError::DatabaseError(e.into()))?
+        .ok_or_else(|| ApiError::InvalidInputError("Invalid Version ID specified!".to_string()))?;
+
+        if team_member
+            .permissions
+            .contains(Permissions::DELETE_VERSION)
+            && team_member.accepted
```
Aeledfyr, line 194: This condition should be removed. `get_from_user_id` can only return accepted team members.

Aeledfyr (changes requested):
I've realized now that `get_from_user_id` can only return accepted team members, so a lot of the `member.accepted == true` checks can be removed.  That means that we have to fix the delete member route though, since that uses `get_from_user_id` to get the member that's being removed, meaning that it can't remove pending requests.

Aeledfyr (approved):

PR #102, opened 2020-11-10
Minor fixes to orderings and permission serialization

Aeledfyr:
* Fixes the local indexer to use the most recently released minecraft version rather than the oldest version
* Fixes the case where a user with invalid permissions would be skipped when listing rather than throwing an error
* Specifies some previously random sort orderings
* Changes `Permissions` to be optional in the `TeamMember` struct and use `#[serde(transparent)]`, making it represented as a normal integer in JSON

### DIFF
--- a/sqlx-data.json
+++ b/sqlx-data.json
@@ -1137,6 +1137,26 @@
       "nullable": []
     }
   },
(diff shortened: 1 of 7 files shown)


Geometrically (approved):

Geometrically:
Did you forget to run prepare?

PR #104, opened 2020-11-23
More mod info

Geometrically:
More mod info, may also add icon/ user editing routes

### DIFF
--- a/.env
+++ b/.env
@@ -35,5 +35,3 @@ INDEX_CACHE_PATH=/tmp/modrinth-id-cache.json
 
 GITHUB_CLIENT_ID=3acffb2e808d16d4b226
 GITHUB_CLIENT_SECRET=none
-
-IP_SALT=labrinth
--- a/.env
+++ b/.env
@@ -35,3 +35,5 @@ INDEX_CACHE_PATH=/tmp/modrinth-id-cache.json
 
 GITHUB_CLIENT_ID=3acffb2e808d16d4b226
 GITHUB_CLIENT_SECRET=none
+
+IP_SALT=labrinth
--- a/migrations/20201122043349_more-mod-data.sql
+++ b/migrations/20201122043349_more-mod-data.sql
@@ -26,6 +26,7 @@ CREATE TABLE side_types (
 INSERT INTO side_types (name) VALUES ('required');
 INSERT INTO side_types (name) VALUES ('no-functionality');
 INSERT INTO side_types (name) VALUES ('unsupported');
+INSERT INTO side_types (name) VALUES ('unknown');
 
 CREATE TABLE licenses (
     id serial PRIMARY KEY,
@@ -33,7 +34,7 @@ CREATE TABLE licenses (
     name varchar(1000) UNIQUE
 );
 
-INSERT INTO licenses (short, name) VALUES ('arr', 'All Rights Reserved');
+INSERT INTO licenses (short, name) VALUES ('custom', 'Custom License');
 
 ALTER TABLE versions
     ADD COLUMN featured BOOLEAN NOT NULL default FALSE;
@@ -45,10 +46,18 @@ ALTER TABLE mods
 ALTER TABLE mods
     ADD COLUMN license_url varchar(1000) NULL;
 ALTER TABLE mods
-    ADD COLUMN client_side integer REFERENCES side_types NOT NULL default 1;
+    ADD COLUMN client_side integer REFERENCES side_types NOT NULL default 4;
 ALTER TABLE mods
-    ADD COLUMN server_side integer REFERENCES side_types NOT NULL default 1;
+    ADD COLUMN server_side integer REFERENCES side_types NOT NULL default 4;
 ALTER TABLE mods
     ADD COLUMN discord_url varchar(255) NULL;
 ALTER TABLE mods
     ADD COLUMN slug varchar(255) NULL UNIQUE;
+
+CREATE TABLE downloads (
+    id serial PRIMARY KEY,
+    version_id bigint REFERENCES versions ON UPDATE CASCADE NOT NULL,
+    date timestamptz DEFAULT CURRENT_TIMESTAMP NOT NULL,
+    -- A SHA1 hash of the downloader IP address
+    identifier varchar(40) NOT NULL
+);
--- a/migrations/20201122043349_more-mod-data.sql
+++ b/migrations/20201122043349_more-mod-data.sql
@@ -1,0 +1,54 @@
+CREATE TABLE donation_platforms (
+    id serial PRIMARY KEY,
+    short varchar(100) UNIQUE,
+    name varchar(500) UNIQUE
+);
+
+INSERT INTO donation_platforms (short, name) VALUES ('patreon', 'Patreon');
+INSERT INTO donation_platforms (short, name) VALUES ('bmac', 'Buy Me a Coffee');
+INSERT INTO donation_platforms (short, name) VALUES ('paypal', 'PayPal');
+INSERT INTO donation_platforms (short, name) VALUES ('github', 'GitHub Sponsors');
+INSERT INTO donation_platforms (short, name) VALUES ('ko-fi', 'Ko-fi');
+INSERT INTO donation_platforms (short, name) VALUES ('other', 'Other');
+
+CREATE TABLE mods_donations (
+    joining_mod_id bigint REFERENCES mods ON UPDATE CASCADE NOT NULL,
+    joining_platform_id int REFERENCES donation_platforms ON UPDATE CASCADE NOT NULL,
+    url varchar(2048) NOT NULL,
+    PRIMARY KEY (joining_mod_id, joining_platform_id)
+);
+
+CREATE TABLE side_types (
+    id serial PRIMARY KEY,
+    name varchar(64) UNIQUE
+);
+
+INSERT INTO side_types (name) VALUES ('required');
+INSERT INTO side_types (name) VALUES ('no-functionality');
+INSERT INTO side_types (name) VALUES ('unsupported');
+
+CREATE TABLE licenses (
+    id serial PRIMARY KEY,
+    short varchar(60) UNIQUE,
+    name varchar(1000) UNIQUE
+);
+
+INSERT INTO licenses (short, name) VALUES ('arr', 'All Rights Reserved');
+
+ALTER TABLE versions
+    ADD COLUMN featured BOOLEAN NOT NULL default FALSE;
+ALTER TABLE files
+    ADD COLUMN "primary" BOOLEAN NOT NULL default FALSE;
+
+ALTER TABLE mods
+    ADD COLUMN license integer REFERENCES licenses NOT NULL default 1;
+ALTER TABLE mods
+    ADD COLUMN license_url varchar(1000) NULL;
+ALTER TABLE mods
+    ADD COLUMN client_side integer REFERENCES side_types NOT NULL default 1;
+ALTER TABLE mods
+    ADD COLUMN server_side integer REFERENCES side_types NOT NULL default 1;
+ALTER TABLE mods
+    ADD COLUMN discord_url varchar(255) NULL;
+ALTER TABLE mods
+    ADD COLUMN slug varchar(255) NULL UNIQUE;
--- a/migrations/20201122043349_more-mod-data.sql
+++ b/migrations/20201122043349_more-mod-data.sql
@@ -1,7 +1,7 @@
 CREATE TABLE donation_platforms (
     id serial PRIMARY KEY,
-    short varchar(100) UNIQUE,
-    name varchar(500) UNIQUE
+    short varchar(100) UNIQUE NOT NULL,
+    name varchar(500) UNIQUE NOT NULL
 );
 
 INSERT INTO donation_platforms (short, name) VALUES ('patreon', 'Patreon');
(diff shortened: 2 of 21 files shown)


migrations/20201122043349_more-mod-data.sql
```diff
@@ -0,0 +1,54 @@
...
+
+INSERT INTO licenses (short, name) VALUES ('arr', 'All Rights Reserved');
+
+ALTER TABLE versions
+    ADD COLUMN featured BOOLEAN NOT NULL default FALSE;
+ALTER TABLE files
+    ADD COLUMN "primary" BOOLEAN NOT NULL default FALSE;
+
+ALTER TABLE mods
+    ADD COLUMN license integer REFERENCES licenses NOT NULL default 1;
```
Aeledfyr, line 44: This should not have a default, or licenses should have an unspecified variant.  This should be required to be specified in the future, but it's not safe to assume that all current mods are all rights reserved.

migrations/20201122043349_more-mod-data.sql
```diff
@@ -0,0 +1,54 @@
...
+ALTER TABLE mods
+    ADD COLUMN license_url varchar(1000) NULL;
+ALTER TABLE mods
+    ADD COLUMN client_side integer REFERENCES side_types NOT NULL default 1;
```
Aeledfyr, line 48: It would also be good to have an unknown variant of this as the default for old mods, that can't be selected when creating a new mod.  That way we know what information is valid and what we have to fix.

src/database/models/mod.rs
```diff
@@ -79,3 +79,69 @@ impl ids::StatusId {
...
+            .fetch_optional(exec)
+            .await?;
+
+        Ok(result.map(|r| ids::SideTypeId(r.id)))
+    }
+}
+
+impl ids::LicenseId {
+    pub async fn get_id<'a, E>(
+        id: &String,
```
Aeledfyr, line 107: This should be `&str`

src/database/models/mod.rs
```diff
@@ -79,3 +79,69 @@ impl ids::StatusId {
...
+            .fetch_optional(exec)
+            .await?;
+
+        Ok(result.map(|r| ids::LicenseId(r.id)))
+    }
+}
+
+impl ids::DonationPlatformId {
+    pub async fn get_id<'a, E>(
+        id: &String,
```
Aeledfyr, line 129: This should be `&str`

src/models/mods.rs
```diff
@@ -35,21 +37,87 @@ pub struct Mod {
...
+pub enum SideType {
+    Required,
+    NoFunctionality,
+    Unsupported,
+    Unknown,
+}
+
+impl std::fmt::Display for SideType {
+    fn fmt(&self, fmt: &mut std::fmt::Formatter) -> std::fmt::Result {
+        match self {
```
Aeledfyr, line 79: This could be reduced down to `write!(fmt, "{}", self.as_str())`

src/routes/mod_creation.rs
```diff
@@ -113,8 +115,20 @@ struct ModCreateData {
...
+    /// An optional link to the mod's discord.
+    pub discord_url: Option<String>,
     /// An optional boolean. If true, the mod will be created as a draft.
     pub is_draft: Option<bool>,
+    /// The support range for the client mod
+    pub client_side: SideType,
+    /// The support range for the server mod
+    pub server_side: SideType,
+    /// The license id that the mod follows
+    pub license_id: String,
```
Aeledfyr, line 129: If licenses are strictly limited to the options we give, there needs to be an "other" category and a way of linking to the license.  People will always come up with weird licenses, so we can't enumerate them all.

src/routes/mod_creation.rs
```diff
@@ -474,22 +516,37 @@ async fn mod_create_inner(
...
             team: team_id.into(),
             title: mod_builder.title.clone(),
             description: mod_builder.description.clone(),
             body_url: mod_builder.body_url.clone(),
             published: now,
             updated: now,
             status,
+            license: License {
+                id: mod_create_data.license_id.clone(),
+                name: "".to_string(),
```
Aeledfyr, line 545: This shouldn't be empty; can you query this when you get the id?

src/routes/mods.rs
```diff
@@ -130,22 +130,33 @@
 fn convert_mod(data: database::models::mod_item::QueryMod) -> models::mods::Mod {
     let m = data.inner;
 
+    // TODO: Fill fields
```
Aeledfyr, line 133: Make sure this gets finished before merging

Aeledfyr (changes requested):
For the licenses, how do you plan on supporting  the large variety of licenses people will use?  How do you plan on supporting dual licensed mods?  We also have to have a way of supporting crayon (custom) licenses.

There's also the question of how to deal with the defaults for mods currently in the system.  We should have an unknown / unspecified variant for each (license and client-/server-side) that is used as the default for old mods, but require new mods to specify a real value.

migrations/20201122043349_more-mod-data.sql
Geometrically: Required is best to be the default, as unknown gives users not much information, while it is safe to assume that every mod is required on the server and client by default.

Geometrically: And the default is only for the data currently in the database

migrations/20201122043349_more-mod-data.sql
Geometrically: It is safe to assume that all mods are ARR, unless specified. It also saves us from legal trouble, because if we specify the wrong license for a mod (on our end), that would be bad for us

src/routes/mod_creation.rs
Geometrically: I'll add a "Custom License" license in the migration for the sake of this

migrations/20201122043349_more-mod-data.sql
```diff
@@ -26,29 +26,38 @@ CREATE TABLE side_types (
...
-    ADD COLUMN slug varchar(255) NULL UNIQUE;
\ No newline at end of file
+    ADD COLUMN slug varchar(255) NULL UNIQUE;
+
+CREATE TABLE downloads (
+    id serial PRIMARY KEY,
+    version_id bigint REFERENCES versions ON UPDATE CASCADE NOT NULL,
+    date timestamptz DEFAULT CURRENT_TIMESTAMP NOT NULL,
+    -- A SHA1 hash of the downloader IP address
+    identifier varchar(40) NOT NULL
```
Aeledfyr, line 62: This is not anonymized, and should probably be deleted after a few days.  Most users will be using IPv4 (Do we even support IPv6? We should), so that gives a bit less than 2^32 addresses - this ends up taking <30 minutes to generate the hash of *every* IPv4 address, allowing this to be entirely de-anonymized.

I'm assuming that this is intended to avoid tracking multiple downloads from the same IP in a short time range?  If so, I  would recommend clearing this field for every record older than an hour.  It would also be good to generate a random salt on startup that is hashed with the IP address, and ensure that that salt is never logged or saved, but that doesn't entirely fix the issue; we need to clear old records on a regular basis.

src/routes/versions.rs
```diff
@@ -520,6 +581,100 @@ pub async fn get_version_from_hash(
     }
 }
 
+#[derive(Serialize, Deserialize)]
+pub struct DownloadRedirect {
+    pub url: String,
+}
+
+// under /api/v1/version_file/{hash}
+#[post("{version_id}")]
```
Aeledfyr, line 590: Why is this a post request?  Shouldn't it be get?

src/routes/moderation.rs
```diff
@@ -54,9 +55,18 @@ pub async fn mods(
             issues_url: m.issues_url,
             source_url: m.source_url,
             status: ModStatus::Processing,
+            license: License {
+                id: "".to_string(),
+                name: "".to_string(),
+                url: None,
+            },
```
Aeledfyr, line 62: This needs to have filled out data for moderation to be usable. This should use the same system as the normal mods route if possible - it will be slower, but you should be able to query a list of ids and then use the exact same code as used in the mods route, so you end up with the full information here.

src/routes/versions.rs
```diff
@@ -518,6 +581,100 @@ pub async fn get_version_from_hash(
...
+    )
+    .fetch_optional(&**pool)
+    .await
+    .map_err(|e| ApiError::DatabaseError(e.into()))?;
+
+    if let Some(id) = result {
+        let download_exists = sqlx::query!(
+            "SELECT EXISTS(SELECT 1 FROM downloads WHERE version_id = $1 AND date > (CURRENT_DATE - INTERVAL '30 minutes ago'))",
+            id.version_id,
+        )
```
Aeledfyr, line 616: This checks whether the mod has been downloaded in the last 30 minutes by *anyone*, not from this specific ip.

src/routes/versions.rs
```diff
@@ -518,6 +581,100 @@ pub async fn get_version_from_hash(
...
+                    $1, $2
+                )
+                ",
+                id.version_id,
+                real_ip
+                    .realip_remote_addr()
+                    .ok_or_else(|| ApiError::InvalidInputError(
+                        "You may only record a download for a mod with a normal client!"
+                            .to_string()
+                    ))?
```
Aeledfyr, line 639: Isn't this supposed to be the sha-1 hash of the IP address?  Also, this should not throw any errors if it can't log it - it should still redirect and let the user download it.

Aeledfyr (changes requested):
There is still some information not being returned by the moderation / other routes (also see previous review comments).  The download tracking system has issues, and isn't great from a privacy standpoint.  The best solution for the privacy issues with the download counting would be to either clear the ip field or just delete the download records for downloads that are more than a few hours old; in addition, hash the ips with a random salt generated at startup.

migrations/20201122043349_more-mod-data.sql
diabolical17: Commenting on the IPv6, its becoming slowly standardised, i suggest using it instead of having to rewrite code in the future. You could use IPv4 as a fallback if IPv6 is unavailable or something 🤷🏽

Geometrically: `realip_remote_addr` supports IPV6 iirc

diabolical17: Ah, nevermind then 😅

src/routes/moderation.rs
Geometrically: This isn't needed, as this is just a list of mods that the moderator needs to approve. When reviewing a mod, a mod will then click on the mods on the frontend and go to the mod's page, which will have this data.

migrations/20201122043349_more-mod-data.sql
Aeledfyr: The IPv6 comment is the fact that our current server hosting stuff doesn't support IPv6.  This is either an issue with the DNS records or the actual host configuration (@Geometrically?)

Geometrically: Wdym? The server isn't on IPV6, but the host should accept IPV6 requests

src/routes/moderation.rs
Aeledfyr: How the *frontend* is structured shouldn't affect the backend design.  If this isn't going to be included, make a new structure for it that doesn't include the data that isn't included.  This route should either return the *exact* same information that the normal mods route does, or return *only* a list of mod ids that need to be reviewed, which would then require sending it to the mods route to get the actual information.

This either has to return the full data, or make a new partial mod data structure that specifies what information is actually being sent instead of filling it in with defaults.

.env
```diff
@@ -34,4 +34,6 @@ EXTERNAL_INDEX_INTERVAL=43200
 INDEX_CACHE_PATH=/tmp/modrinth-id-cache.json
 
 GITHUB_CLIENT_ID=3acffb2e808d16d4b226
-GITHUB_CLIENT_SECRET=none
\ No newline at end of file
+GITHUB_CLIENT_SECRET=none
+
+IP_SALT=labrinth
```
Aeledfyr, line 39: This isn't what I was talking about; this needs to be generated randomly on server startup, or it isn't useful at all.  If it's static, even if it's configured to be something else on the production host, then it is just as easy to brute force as without it.

This isn't as much of an issue with older downloads being cleared, but this doesn't do much as is - it should be easy to add a random number to the end on startup and then pass that to the routes using actix's shared data.

src/database/models/categories.rs
```diff
@@ -453,3 +465,297 @@ impl<'a> GameVersionBuilder<'a> {
...
+            // Nothing was deleted
+            Ok(None)
+        } else {
+            Ok(Some(()))
+        }
+    }
+}
+
+impl<'a> LicenseBuilder<'a> {
+    /// The game version.  Spaces must be replaced with '_' for it to be valid
```
Aeledfyr, line 567: Copy-pasted docs, go through and fix the other copies of this - also in the insert method.

Aeledfyr (changes requested):
A few things - the moderation routes need to return the same data as the normal mod routes, and the salt currently isn't implemented properly.  (Also, the build is failing.  Did you run `cargo sqlx prepare`?)

Aeledfyr (approved):
Don't merge this until the frontend supports the changes to mod creation, but it looks like it should be ready to go.

PR #106, opened 2020-11-28
Fix hash

Geometrically:
High priority fix - hash lookup in database does NOT work correctly, this PR fixes the SQL query used.

It also makes users not able to see versions they are unauthorized to see.


src/routes/versions.rs
```diff
@@ -703,15 +711,7 @@ pub async fn delete_file(
...
-        let version_data = database::models::Version::get_full(
-            database::models::VersionId(row.version_id),
-            &**pool,
-        )
-        .await
-        .map_err(|e| ApiError::DatabaseError(e.into()))?;
-
-        if let Some(data) = version_data {
-            let mut transaction = pool
+         let mut transaction = pool
```
Aeledfyr, line 714: 1. The indentation is messed up past this point, run cargo fmt and fix the sql query strings.
2. Wasn't this verifying permissions somehow? How is it doing that now?

Aeledfyr (commented):
Have you tested this? It needs cargo fmt and testing to make sure the sql queries actually work.

Geometrically:
I tested all the queries. Running formatter once I'm home.

Geometrically:
The majority of the diff is from the other PR. I forgot to base this PR off of master

src/routes/versions.rs
Geometrically: It still checks them in the query

Aeledfyr (approved):

PR #109, opened 2020-12-02
Fix access controls

Geometrically:


src/routes/mod_creation.rs
```diff
@@ -582,11 +579,6 @@ async fn mod_create_inner(
 
         let _mod_id = mod_builder.insert(&mut *transaction).await?;
 
-        let index_mod =
-            crate::search::indexing::local_import::query_one(mod_id.into(), &mut *transaction)
-                .await?;
-        indexing_queue.add(index_mod);
```
Aeledfyr, line 588: Instead of removing this, this should check whether the mod's initial status should be visible, like the other local indexer does.  This makes it more consistent, and won't end up with mods having to wait 15 minutes when they should have appeared immediately.

This would only show up if we allow mods to automatically show up in the search results before moderation, but having this would allow us to make that change with fewer bugs; it's also more consistent with what the other indexer does.

Aeledfyr (changes requested):

src/routes/mod_creation.rs
Geometrically: There is no initial status that makes it visible

Geometrically: I will add to moderation routes to add it to indexing if the new status is searchable

Aeledfyr (approved):
I'm not a fan of fully removing Curseforge indexing rather than disabling it, but I'm okay with it if it is necessary for legal or similar reasons (such as not being banned from Fabric communities).

PR #119, opened 2020-12-27
Add 'accepted' field to TeamMember

Geometrically:
This PR makes it so the frontend can check if a team member is accepted or not. It also makes it so routes with authorization returns unaccepted members.

### DIFF
--- a/src/routes/teams.rs
+++ b/src/routes/teams.rs
@@ -30,6 +30,7 @@ pub async fn team_members_get(
                     user_id: data.user_id.into(),
                     role: data.role,
                     permissions: Some(data.permissions),
+                    accepted: data.accepted,
                 })
                 .collect();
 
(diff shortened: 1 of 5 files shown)


src/routes/teams.rs
```diff
@@ -30,21 +30,28 @@ pub async fn team_members_get(
...
                 })
                 .collect();
 
             return Ok(HttpResponse::Ok().json(team_members));
         }
     }
 
-    let team_members: Vec<crate::models::teams::TeamMember> = members_data
-        .into_iter()
-        .map(|data| crate::models::teams::TeamMember {
```
Aeledfyr, line 42: You could just add `.filter(|m| team_member.accepted)` here and not have to change anything else

Aeledfyr (approved):

Mysterious-Dev (approved):

src/routes/teams.rs
Geometrically: Hmm okay, will do that in the future. I'm a bit used to Linq and that causing double iteration.

PR #143, opened 2021-01-25
Query Optimization

Geometrically:
This PR eliminates the query spam when requesting multiple of an item, or retrieving the full versions of the Version struct and the Mod struct.


Geometrically:
I tested, sorry code reviewers :P

PR #160, opened 2021-02-07
Add Twitter and Github field for user

Mysterious-Dev:
This PR add Twitter and Github field to user database to add social link after on frontend.


Geometrically:
Thank you for the PR, however I'm not sure if this is needed at the moment, I can keep it open for now but I was thinking of approaching this differently for dynamic integrations.

PR #168, opened 2021-02-18
Add check to prevent collisions between slugs and mod IDs

### LINKED ISSUE #167: Slug can be set to other mod's id
**Describe the bug**
A mod's slug can be set to another mod's id, causing links to the former to show the page of the latter. This includes links in search and in the dashboard (excluding the link to mod settings). This means that you can create a mod with a description and icon of your choice and link to a mod that you don't own.

**To Reproduce**
1. Change mod slug to id of other mod
2. Click on link to first mod
3. See that the page of the second mod appears

**Expected behavior**
The mod editing endpoint should error if the new slug collides with another mod's id.

BasiqueEvangelist:
Fixes #167

### DIFF
--- a/src/routes/mod_creation.rs
+++ b/src/routes/mod_creation.rs
@@ -46,6 +46,8 @@ pub enum CreateError {
     InvalidCategory(String),
     #[error("Invalid file type for version file: {0}")]
     InvalidFileType(String),
+    #[error("Slug collides with other mod's id!")]
+    SlugCollision,
     #[error("Authentication Error: {0}")]
     Unauthorized(#[from] AuthenticationError),
     #[error("Authentication Error: {0}")]
@@ -71,6 +73,7 @@ impl actix_web::ResponseError for CreateError {
             CreateError::InvalidFileType(..) => StatusCode::BAD_REQUEST,
             CreateError::Unauthorized(..) => StatusCode::UNAUTHORIZED,
             CreateError::CustomAuthenticationError(..) => StatusCode::UNAUTHORIZED,
+            CreateError::SlugCollision => StatusCode::BAD_REQUEST,
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/routes/mod.rs
```diff
@@ -107,6 +107,8 @@ pub enum ApiError {
     CustomAuthenticationError(String),
     #[error("Invalid Input: {0}")]
     InvalidInputError(String),
+    #[error("Validation Error: {0}")]
+    ValidationError(String),
```
Geometrically, line 111: Don't create a new error type, I think `InvalidInputError` works fine for this case.

src/routes/mod.rs
```diff
@@ -125,6 +127,7 @@ impl actix_web::ResponseError for ApiError {
             ApiError::IndexingError(..) => actix_web::http::StatusCode::INTERNAL_SERVER_ERROR,
             ApiError::FileHostingError(..) => actix_web::http::StatusCode::INTERNAL_SERVER_ERROR,
             ApiError::InvalidInputError(..) => actix_web::http::StatusCode::BAD_REQUEST,
+            ApiError::ValidationError(..) => actix_web::http::StatusCode::BAD_REQUEST,
```
Geometrically, line 130: See above

src/routes/mod.rs
```diff
@@ -141,6 +144,7 @@ impl actix_web::ResponseError for ApiError {
                     ApiError::IndexingError(..) => "indexing_error",
                     ApiError::FileHostingError(..) => "file_hosting_error",
                     ApiError::InvalidInputError(..) => "invalid_input",
+                    ApiError::ValidationError(..) => "validation_error",
```
Geometrically, line 147: See above

src/routes/mod_creation.rs
```diff
@@ -498,6 +502,25 @@ async fn mod_create_inner(
...
+            )
+            .fetch_one(&mut *transaction)
+            .await
+            .map_err(|e| CreateError::DatabaseError(e.into()))?;
+
+            if results.exists.unwrap_or(true) {
+                return Err(CreateError::SlugCollision);
+            }
+        }
+
```
Geometrically, line 523: Make this higher up where all the other validation stuff is, that way files aren't uploaded and deleted, which is a waste of bandwidth and processing.

Geometrically (changes requested):

Geometrically (approved):

PR #180, opened 2021-03-28
Maven endpoint support

BasiqueEvangelist:
This allows mod developers to add the Modrinth API as a maven repository and depend on mods hosted on Modrinth.


src/database/models/mod_item.rs
```diff
@@ -405,7 +405,7 @@ impl Mod {
     }
 
     pub async fn get_full_from_slug<'a, 'b, E>(
-        slug: String,
+        slug: &String,
```
Redblueflame, line 408: Why use a String borrow?
In rust, variable can't be copied without any information, unless they derive the Copy trait.
In our case, string does not derive Copy, but it derives Clone, a verbose copy, meaning that the item is copied only when you call a function, `x.clone()`. In all cases where the string is needed, but you only have a borrow, clone it.

src/routes/maven.rs
```diff
@@ -0,0 +1,262 @@
...
+    req: HttpRequest,
+    info: web::Path<(String,)>,
+    pool: web::Data<PgPool>,
+) -> Result<HttpResponse, ApiError> {
+    let string = info.into_inner().0;
+    let id_option: Option<ModId> = serde_json::from_str(&*format!("\"{}\"", string)).ok();
+
+    let mut mod_data;
+
+    if let Some(id) = id_option {
```
Redblueflame, line 62: It's great that you used the if let pattern, it makes it really easier to read

src/routes/maven.rs
```diff
@@ -0,0 +1,262 @@
...
+        mod_data = database::models::Mod::get_full(id.into(), &**pool)
+            .await
+            .map_err(|e| ApiError::DatabaseError(e.into()))?;
```
Redblueflame, line 65: If I recall correctly, we are using the `thiserror` crate, that allows us to skip entirely this mapping process, so you should be able to use the `?` shorthand directly after the await.

src/routes/maven.rs
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    let user_option = get_user_from_headers(req.headers(), &**pool).await.ok();
+
+    if let Some(data) = mod_data {
+        let mut authorized = !data.status.is_hidden();
+
+        if let Some(user) = user_option {
```
Redblueflame, line 83: That kind of functions should have its own function, as it's really possible ti will be needed somewhere else in the code, and will also allow us to reduce the identation in this part of the code.

src/routes/maven.rs
```diff
@@ -0,0 +1,262 @@
...
+                    },
+                    last_updated: data.inner.updated.format("%Y%m%d%H%M%S").to_string(),
+                },
+            };
+
+            // return Ok(HttpResponse::Ok().content_type("text/xml").body(response));
+            Ok(HttpResponse::Ok()
+                .content_type("text/xml")
+                .body(yaserde::ser::to_string(&respdata).map_err(|e| ApiError::XmlError(e))?))
+        } else {
```
Redblueflame, line 146: Rust allows for early returns, so an easy fix for that nesting would be to invert the conditions, and just return the `HttpResponse::NotFound` error.

src/routes/maven.rs
```diff
@@ -0,0 +1,262 @@
...
+pub async fn version_file(
+    req: HttpRequest,
+    web::Path((string, vnum, file)): web::Path<(String, String, String)>,
+    pool: web::Data<PgPool>,
+) -> Result<HttpResponse, ApiError> {
+    let id_option: Option<ModId> = serde_json::from_str(&*format!("\"{}\"", string)).ok();
+
+    let mut mod_data;
+
+    if let Some(id) = id_option {
```
Redblueflame, line 164: You do not even need to have a mutable variable here, rust allows this kind of code to run:
```rust
let data = if request.is_ok() {
  request.data
} else {
  Data::new()
}
```

src/routes/maven.rs
```diff
@@ -0,0 +1,262 @@
...
+    } else {
+        mod_data = database::models::Mod::get_full_from_slug(&string, &**pool)
+            .await
+            .map_err(|e| ApiError::DatabaseError(e.into()))?;
+    }
+
+    let user_option = get_user_from_headers(req.headers(), &**pool).await.ok();
+
+    if let Some(data) = mod_data {
+        let mut authorized = !data.status.is_hidden();
```
Redblueflame, line 183: It's reused here, so making an independant function will be useful.

src/routes/maven.rs
```diff
@@ -0,0 +1,262 @@
...
+                    .map_err(|e| ApiError::DatabaseError(e.into()))?
+                    .exists;
+
+                    authorized = mod_exists.unwrap_or(false);
+                }
+            }
+        }
+
+        if authorized {
+            let vid_option = sqlx::query!(
```
Redblueflame, line 208: One thing that is more dependant on code rot, but when I originally put out the file format, things like requests to the database were meant to be on the models directory with the actual model struct.

Just know that I may change that in the future, and that will make the code far more readable.

Redblueflame (changes requested):
Here is the review, with some insights into how to clean the code, but other than that, it seems really good with how the difficult / inreadable the maven format is.

src/database/models/mod_item.rs
BasiqueEvangelist: Oh.
I actually wanted to replace the String move with a `&str` borrow, but why clone the String?

Redblueflame: The &str borrow is quite difficult to do, mainly with the lifetime management.
Afterall, it's just a preference, and by convention, only &str and String are allowed

src/routes/maven.rs
BasiqueEvangelist: I can't just use ? here because the `sqlx::error::Error` is turned into a `labrinth::database::models::DatabaseError` using the `?` operator, and then is wrapped by an `ApiError`. The `?` operator can't be used alone here because there isn't any direct conversion from `sqlx::error::Error` to `ApiError`.

Redblueflame: In that case, even if it's not the best looking thing, you could implement From<sqlx::error::Error> for ApiError that wraps everything in place.
It should be a 5 line thing, and that will make that part of the code far more readable.

src/database/models/mod_item.rs
BasiqueEvangelist: I replaced the &String borrow with a &str borrow here, as lifetimes really don't matter here (the slug just goes to the sqlx query and that's it)

Redblueflame: Perfect then

Redblueflame (approved):
Looks good to me!
Thanks for your contribution :)

PR #203, opened 2021-05-29
Added caching and fixed typo

Redblueflame:


src/database/cache/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+                    Mutex::new(SizedCache::with_size(400));
+            }
+
+            pub async fn $getter_name<'a>(id: $id) -> Option<$val> {
+                let mut cache = $cache_name.lock().await;
+                info!("Querying cache {} for {}", stringify!($name), id);
+                match Cached::cache_get(&mut *cache, &id) {
+                    Some(e) => Some(e.clone()),
+                    None => None,
+                }
```
Geometrically, line 22: Might want to fix the clippy warnings

src/database/cache/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+            use lazy_static::lazy_static;
+            use log::info;
+            lazy_static! {
+                static ref $cache_name: Mutex<SizedCache<$id, $val>> =
+                    Mutex::new(SizedCache::with_size(400));
+            }
+
+            pub async fn $getter_name<'a>(id: $id) -> Option<$val> {
+                let mut cache = $cache_name.lock().await;
+                info!("Querying cache {} for {}", stringify!($name), id);
```
Geometrically, line 18: Info calls are enabled even with debig false, right? Is there a log debug call?

Geometrically (approved):

src/database/cache/mod.rs
Redblueflame: Currently, labrinth lacks a way to control the log levels in the code, as displayed in the line present in the main.rs file:
```rust
env_logger::Builder::from_env(Env::default().default_filter_or("info")).init();
```
So there is a debug macro in the log crate, but we won't be able to get them to display.

PR #212, opened 2021-06-14
Final V2 Changes

Geometrically:
This PR:
- Moves most database operations to transactions
- Refactors dependencies
- Adds rejection reasons
- Adds an archived status
- Fix Icon Caching issue 
- Moderation Discord Webhook
- Delete multiple notifications route 
- Easy serialization for validation errors
- Make it required to create a version before submitting a project for review 
- Fix notification messages


migrations/20210611024943_archived-status-notifications-icon-rejection-reasons.sql
```diff
@@ -0,0 +1,18 @@
+INSERT INTO statuses (status) VALUES ('archived');
+
+ALTER TABLE notifications
+    ADD COLUMN icon varchar(20000);
+
+ALTER TABLE mods
```
Redblueflame, line 6: This is fine for now, but we should really move to a ticketing system for rejections, as we usually want to get a response instead of blindly refusing it.

src/database/models/version_item.rs
```diff
@@ -10,13 +10,62 @@ pub struct VersionBuilder {
...
+                version_id as VersionId,
+            )
+                .fetch_optional(&mut *transaction).await?.map(|x| VersionId(x.id))
+        } else {
+            self.version_id
+        };
+
+        sqlx::query!(
+            "
+            INSERT INTO dependencies (dependent_id, dependency_type, dependency_id, mod_dependency_id)
```
Redblueflame, line 54: For better readability, wouldn't it better to name it `dependency_version_id` and `dependency_mod_id`?

On another note, I still think that we may need to have a semver like system for dependencies, as it will allow way more flexibility for the dependency resolver.

src/database/models/version_item.rs
```diff
@@ -311,52 +415,61 @@ impl Version {
...
         )
-        .execute(exec)
+        .execute(&mut *transaction)
         .await?;
 
-        sqlx::query!(
+        // Sync dependencies
+
+        let project_id = sqlx::query!(
             "
```
Redblueflame, line 424: Github, you're concerning

src/main.rs
```diff
@@ -277,16 +277,16 @@ async fn main() -> std::io::Result<()> {
...
+                            // (300 requests per second)
+                            let random = rand::thread_rng().gen_range(1, 30);
                             return Ok(format!("{}-{}", ip, random));
                         }
 
                         Ok(ip)
                     })
                     .with_interval(std::time::Duration::from_secs(60))
-                    .with_max_requests(200),
+                    .with_max_requests(600),
```
Redblueflame, line 289: I'ld advocate against upping the request amounts, and adding support should be relatively easy as we use cloudflare, who adds a `CF-Connecting-IP` header with the connecting user id.

src/routes/auth.rs
```diff
@@ -129,78 +129,82 @@ pub async fn auth_callback(
...
             ",
         state_id as i64
     )
-    .fetch_one(&mut *transaction)
+    .fetch_optional(&mut *transaction)
     .await?;
 
-    let now = Utc::now();
-    let duration = result.expires.signed_duration_since(now);
+    if let Some(result) = result_option {
```
Redblueflame, line 142: Invert that condition, you can make the code way more readable by testing for the contrary, and returning an error early.

src/routes/moderation.rs
```diff
@@ -51,3 +52,103 @@ pub async fn get_projects(
 
     Ok(HttpResponse::Ok().json(projects))
 }
+
```
Redblueflame, line 55: If we provide a way for modders to put custom webhooks for releases, wouldn't it be better to move this to a utility file ?
This may also help to reduce the length of this file.

src/routes/project_creation.rs
```diff
@@ -765,3 +785,47 @@ pub fn get_image_content_type(extension: &str) -> Option<&'static str> {
         None
     }
 }
+
+//TODO: In order to ensure readability, only the first error is printed, this may need to be expanded on in the future!
+pub fn validation_errors_to_string(errors: ValidationErrors, adder: Option<String>) -> String {
```
Redblueflame, line 790: Same thing, move it to an utility class, as we may also need this for versions and other things in the future.

src/routes/projects.rs
```diff
@@ -91,7 +94,8 @@ pub async fn project_get(
     let string = info.into_inner().0;
 
     let project_data =
-        database::models::Project::get_full_from_slug_or_project_id(string, &**pool).await?;
+        database::models::Project::get_full_from_slug_or_project_id(string.clone(), &**pool)
```
Redblueflame, line 97: Did cargo-fmt do that?

src/routes/projects.rs
```diff
@@ -146,6 +150,14 @@ pub fn convert_project(
         published: m.published,
         updated: m.updated,
         status: data.status,
+        rejection_data: if let Some(reason) = m.rejection_reason {
```
Redblueflame, line 153: Did modrinth already support rejection reasons in the legacy projects?
  It looks useless if no moderator used it.

src/routes/projects.rs
```diff
@@ -176,7 +188,7 @@ pub fn convert_project(
 }
 
 lazy_static! {
-    static ref RE_URL_SAFE: Regex = Regex::new(r"^[a-zA-Z0-9_-]*$").unwrap();
+    static ref RE_URL_SAFE: Regex = Regex::new(r"\S").unwrap();
```
Redblueflame, line 191: This checks for non space characters in string, not the alpha-numeric-ness of the string. Please revert this change.

src/routes/projects.rs
```diff
@@ -238,6 +249,22 @@ pub struct EditProject {
...
     pub slug: Option<Option<String>>,
+
+    pub status: Option<ProjectStatus>,
+    #[serde(
+        default,
+        skip_serializing_if = "Option::is_none",
+        with = "::serde_with::rust::double_option"
+    )]
+    #[validate(length(max = 2000))]
+    pub rejection_reason: Option<Option<String>>,
```
Redblueflame, line 260: Double option :concern:

src/routes/projects.rs
```diff
@@ -782,12 +867,14 @@ pub async fn project_icon_edit(
...
 
+        let hash = sha1::Sha1::from(bytes.clone()).hexdigest();
+
         let project_id: ProjectId = project_item.id.into();
 
         let upload_data = file_host
             .upload_file(
                 content_type,
-                &format!("data/{}/icon.{}", project_id, ext.ext),
+                &format!("data/{}/{}.{}", project_id, hash, ext.ext),
```
Redblueflame, line 877: Finally 🎉

src/routes/projects.rs
```diff
@@ -851,7 +938,11 @@ pub async fn project_delete(
         }
     }
 
-    let result = database::models::Project::remove_full(project.id, &**pool).await?;
+    let mut transaction = pool.begin().await?;
+
+    let result = database::models::Project::remove_full(project.id, &mut transaction).await?;
```
Redblueflame, line 943: Also remove from cache

src/routes/projects.rs
```diff
@@ -782,12 +867,14 @@ pub async fn project_icon_edit(
...
         let project_id: ProjectId = project_item.id.into();
 
         let upload_data = file_host
             .upload_file(
                 content_type,
-                &format!("data/{}/icon.{}", project_id, ext.ext),
+                &format!("data/{}/{}.{}", project_id, hash, ext.ext),
                 bytes.to_vec(),
             )
             .await?;
```
Redblueflame, line 880: Also remove the element from the cache.

src/routes/teams.rs
```diff
@@ -246,6 +246,11 @@ pub async fn add_team_member(
 
     let team: TeamId = team_id.into();
     NotificationBuilder {
+        icon: Some(
```
Redblueflame, line 249: Why hardcoded? Please host this on the cdn, so that this is more stateless.

src/routes/version_creation.rs
```diff
@@ -20,7 +20,7 @@ use sqlx::postgres::PgPool;
 use validator::Validate;
 
 lazy_static! {
-    static ref RE_URL_SAFE: Regex = Regex::new(r"^[a-zA-Z0-9_\-.]*$").unwrap();
+    static ref RE_URL_SAFE: Regex = Regex::new(r"\S").unwrap();
```
Redblueflame, line 23: Again, this is not URL SAFE;
This is the second time I see the same regex in a lazy static, would it be better to move this to a utility class with all of the more common regexes?

src/routes/version_creation.rs
```diff
@@ -332,9 +338,14 @@ async fn version_create_inner(
     let version_id: VersionId = builder.version_id.into();
 
     NotificationBuilder {
-        title: "A project you followed has been updated!".to_string(),
+        icon: Some(
```
Redblueflame, line 341: Again, CDN  :P

Redblueflame (changes requested):
The PR looks good, but there are some things to change to make the code more readable & fully functional.

src/database/models/version_item.rs
Geometrically: Semver isn't possible, since most mods on Modrinth do not follow semver. It's not called dependency_version_id since it doesn't work that way - I can kinda explain it in discord. If dependency_id is not present, it's auto specified. It is very rare for it to be null

migrations/20210611024943_archived-status-notifications-icon-rejection-reasons.sql
Geometrically: Agreed. It's a temporary solution while we make a robust solution that will be permanent

src/main.rs
Geometrically: I'll change this and the downloads code to use the CF-Connecting-IP

src/routes/projects.rs
Geometrically: No

src/routes/projects.rs
Geometrically: Double options are used to remove the rejection reason. Speaking of which, I should make it so when the user sends the mod for approval the rejection reason is removed

src/routes/projects.rs
Geometrically: yes

src/database/models/version_item.rs
Redblueflame: That didn't answer the possible rename for better clarity

Redblueflame (approved):
Looks good to me, there are some Clippy issues that you may fix before merging tho.

PR #225, opened 2021-07-27
Detect if redirect url contains a query string

venashial:
Currently, when auth redirects a user back to the frontend, it doesn't account for situations where the redirect URL already includes a query string. This results in issues such as the mod search page interpreting the code as a search term:
image


wafflecoffee:
Hope that isn't an actual auth code! :grimacing:

Geometrically (approved):
LGTM

PR #238, opened 2021-08-23
Add purpur support migration

Redblueflame:
This PR is related to modrinth/knossos#301
ANd is created to add support of purpur in the staging server.


Geometrically:
I think it's best to handle this as a new project type (Plugins) and have the loaders be Purpur + Bukkit/Spigot. This would require a bit more work code-wise, but it would work well with the frontend code.

PR #240, opened 2021-08-26
Added support for compatible loaders & add default loaders to dev environments

Redblueflame:
This PR was created in order to create a way to display information about what plugins are supported on what platform.
This also includes a way for mod loaders to be superseded (this may be useful in the case of quilt.

What still needs to be done:

- [ ] Add support for it in the search (add the corresponding loader category for superseding loader-using mods)
- [ ] Add proxy plugins, like bungeecord, as it does have a place within the plugins, but may need another category filter to work at its best.
- [ ] Add the paper logo, as it had issues when I first got it, waiting for an updated version, I will upload soon.


Redblueflame:
Update: Support for proxy won't be available as of the v2 release, but will be added later.

Redblueflame:
This PR should be ready to review and test.

src/validate/sponge.rs
```diff
@@ -0,0 +1,37 @@
...
+        archive: &mut ZipArchive<Cursor<&[u8]>>,
+    ) -> Result<ValidationResult, ValidationError> {
+        // TODO: Add schema validation in order to make sure that it won't error out
+        archive.by_name("mcmod.info").map_err(|_| {
+            ValidationError::InvalidInputError(
+                "No mcmod.info file is present in your file.".to_string(),
+            )
+        })?;
+
+        Ok(ValidationResult::Pass)
```
Geometrically, line 35: Is this a mistake? If I recall correctly Sponge JARs don't have an mcmod.info

src/database/models/categories.rs
```diff
@@ -303,11 +304,14 @@ impl Loader {
...
             ARRAY_AGG(DISTINCT pt.name) project_types
             FROM loaders l
-            LEFT OUTER JOIN loaders_project_types lpt ON joining_loader_id = l.id
-            LEFT OUTER JOIN project_types pt ON lpt.joining_project_type_id = pt.id
-            GROUP BY l.id;
+                     LEFT OUTER JOIN loaders_project_types lpt ON joining_loader_id = l.id
+                     LEFT OUTER JOIN project_types pt ON lpt.joining_project_type_id = pt.id
+                     LEFT OUTER JOIN loaders_relations lr on l.id = lr.child_loader
+                     LEFT OUTER JOIN loaders lrt on lr.parent_loader = lrt.id
+            GROUP BY l.id
```
Geometrically, line 314: Avoid LEFT JOINs (they have a huge peformance cost). You can look at other examples in the codebase where they were replaced for multiple queries.

src/database/models/categories.rs
```diff
@@ -12,6 +12,7 @@ pub struct Loader {
     pub loader: String,
     pub icon: String,
     pub supported_project_types: Vec<String>,
+    pub supported_loaders: Vec<String>,
```
Geometrically, line 15: This name is kinda off. I think "child_loaders" would work better. I think child loaders should also work with search - essentially, if a loader has child loaders, all of those loaders are added to the search document's loaders.

Geometrically (changes requested):
LGTM, but needs some minor changes

src/validate/sponge.rs
Redblueflame: https://github.com/SpongePowered/plugin-meta/blob/0.8.x/mcmod-info/src/test/resources/org/spongepowered/plugin/meta/mcmod1.info.json
They have

src/database/models/categories.rs
Redblueflame: The performance impact in this specific case is really low (the number of records won't be really high), and in order to make this performant, caching may be needed for that.

deadmeu:
You may wish to reduce the size of the svg logos using [svgo](https://github.com/svg/svgo). For example, I was able to reduce the size of the Forge logo by 31%.

wafflecoffee:
Superseded by recent commits.

PR #248, opened 2021-09-12
Remove contributing file and instead direct people to docs

wafflecoffee:
This fixes a long standing issue where people would point to the frontend homepage where it says 'fully documented' then point to the README here and say 'well that's sure hypocritical'

Thus the branch name :tiny_pumpkin:

[no ci]


Redblueflame (approved):
LGTM

PR #252, opened 2021-10-01
Maven repo should return primary file by default

Kir-Antipov:
The title is pretty much self-descriptive - maven repository should return primary file for `{id}-{version}.jar` queries. At the moment it returns last file provided by the database, which can be pretty random:

https://github.com/modrinth/labrinth/blob/91065a616893dbc1c19447057fce3c0b4b59fc21/src/database/models/version_item.rs#L634-L636

As you can see, no `ORDER BY` is specified, therefore we **cannot** rely on the order.

### DIFF
--- a/src/routes/maven.rs
+++ b/src/routes/maven.rs
@@ -234,6 +234,10 @@ pub async fn version_file(
             return Ok(HttpResponse::TemporaryRedirect()
                 .header("Location", &*selected_file.url)
                 .body(""));
+        } else if let Some(selected_file) = version.files.iter().last() {
+            return Ok(HttpResponse::TemporaryRedirect()
+                .header("Location", &*selected_file.url)
+                .body(""));
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/routes/maven.rs
```diff
@@ -230,7 +230,7 @@ pub async fn version_file(
             .header("Location", &*selected_file.url)
             .body(""));
     } else if file == format!("{}-{}.jar", &string, &version.version_number) {
-        if let Some(selected_file) = version.files.iter().last() {
+        if let Some(selected_file) = version.files.iter().find(|x| x.primary) {
```
Geometrically, line 233: Not all versions on modrinth have a primary file. Instead of erroring when the version has no primary file, the behavior should be to return the last file of the version. Could you create a new if statement which would handle this case?

Geometrically (approved):

src/routes/maven.rs
Kir-Antipov: Oh, ok. I thought that all versions have a primary one. Give me a moment

Kir-Antipov: Done. If you want me to rewrite the thing within a single if - let me know

PR #306, opened 2022-02-22
Move download counting to worker

enderger:
Moves download counting to be called by a CloudFlare worker. Needs to be tested on `staging` due to overhead of testing without a frontend, might make some kind of programme (or EMACS major mode) to allow easy local testing.

### DIFF
--- a/.env
+++ b/.env
@@ -3,6 +3,7 @@ RUST_LOG=info,sqlx::query=warn
 
 SITE_URL=https://modrinth.com
 CDN_URL=https://cdn.modrinth.com
+LABRINTH_ADMIN_KEY=feedbeef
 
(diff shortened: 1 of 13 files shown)


enderger:
> Wouldn't it be a lot easier if this used `util::auth::check_is_admin_from_headers`?

In this case, it's not user-based. This is intended to be used as a route for a CloudFlare worker in order to count downloads, and is hence separate.

Geometrically (changes requested):
Looks good! I would run `cargo fmt` and `cargo clippy` to format and solve some warnings (mainly unused vars)

Geometrically (approved):
LGTM!

PR #335, opened 2022-04-19
Allow icons to be 1MiB instead of 256KiB

wafflecoffee:


Geometrically:
I am kind of split about this. 1 MB is a lot, especially on mobile devices and slower connections. I think it's better to just do some knossos compression on the client side.

PR #353, opened 2022-05-24
Fix wrongfully parsed hashes for the version_files route(s)

4JX:
Before it would actually hash the bytes of the hash string, leading to the erroneous hashes.

Not too sure about just unwrapping there, but it technically should never fail

Example response now:
yay


wafflecoffee:
I think you can also remove the `hex` crate now

wafflecoffee (approved):

Geometrically (changes requested):
No unwraps. All errors must be handled. Look at ApiError and give it an appropriate error

4JX:
As in assigning it an `ApiError`? Or do you mean to take that as a reference and using a locally declared one like `AuthorizationError` does?

If its the former I'm kind of lost at where to put it, since a "db hash is invalid" error would need an oddly specific enum entry in that list.

Geometrically:
> As in putting it directly in `ApiError`? Or do you mean to take that as a reference and using a locally declared one like `AuthorizationError` does?
> 
> If its the former I'm kind of lost at where to put it, since a "db hash (string) is invalid" error would need an oddly specific enum entry in that list. Or would something like [this](https://github.com/4JX/labrinth/commit/6beccecbdbf23fffefd29c28d500c16e0a5441b5) work? (Figured I'd leave the error string generic, since you may not always have a `version_id` at hand if this happens somewhere else).

If it's a database error, instantiate `DatabaseError::Other` and convert it to an API error. If it's a user, use `APIError::InvalidInput`

4JX:
> If it's a database error, instantiate `DatabaseError::Other` and convert it to an API error. If it's a user, use `APIError::InvalidInput`
 
Ouch, somehow didn't notice that `Database` referred to a local struct and not an imported crate. Of note is that the routes still return a 200 regardless of whether anything was actually found, may I extend the PR a little to cover that?

Geometrically:
> > If it's a database error, instantiate `DatabaseError::Other` and convert it to an API error. If it's a user, use `APIError::InvalidInput`
> 
> Ouch, somehow didn't notice that `Database` referred to a local struct and not an imported crate. Of note is that the routes still return a 200 regardless of whether anything was actually found, may I extend the PR a little to cover that?

It shouldn't return a 404, the 200 is fine b/c the response is an empty array

Geometrically (approved):

Geometrically:
Thanks for the PR!

PR #355, opened 2022-05-25
Always return the username as fetched from the database

### LINKED ISSUE #350: Name displayed on profile page is affected by the entered url
**Describe the bug**
https://modrinth.com/user/Azim
https://modrinth.com/user/azim
https://modrinth.com/user/aZiM
while they show the same profile, the name shown under the avatar changes depending on the url

**To Reproduce**
1. Go to any user page
2. Play around with username in the url
3. See error

**Expected behavior**
Exact username to be shown no matter the url's case

**Additional context**
`Go to my profile` uses lowercase version of username, and trying to change the username in settings gives `Username .. is taken!` error, even though its technically same profile

itsCryne:
Fixes #350

### DIFF
--- a/src/scheduler.rs
+++ b/src/scheduler.rs
@@ -96,7 +96,7 @@ async fn update_versions(
     pool: &sqlx::Pool<sqlx::Postgres>,
 ) -> Result<(), VersionIndexingError> {
     let input = reqwest::get(
-        "https://launchermeta.mojang.com/mc/game/version_manifest.json",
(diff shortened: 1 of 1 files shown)


Geometrically (changes requested):
Your PR has substantial changes (which are not needed) to the sqlx-data. Please reset it and regenerate it with `cargo sqlx prepare`

wafflecoffee:
whoops, how did I manage to do that?

PR #373, opened 2022-06-17
Adds /teams route

Ricky12Awesome:
Implements #331

### DIFF
--- a/src/routes/teams.rs
+++ b/src/routes/teams.rs
@@ -1,7 +1,7 @@
 use crate::database::models::notification_item::{
     NotificationActionBuilder, NotificationBuilder,
 };
-use crate::database::models::{Team, TeamMember};
+use crate::database::models::TeamMember;
 use crate::models::ids::ProjectId;
 use crate::models::teams::{Permissions, TeamId};
 use crate::models::users::UserId;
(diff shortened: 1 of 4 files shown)


wafflecoffee:
You're going to need to run `cargo sqlx prepare` for it to compile with GitHub Actions.

wafflecoffee (approved):

src/routes/teams.rs
```diff
@@ -101,6 +101,65 @@ pub async fn team_members_get(
...
+    let teams_data =
+        TeamMember::get_from_team_full_many(team_ids, &**pool).await?;
+
+    let teams_groups = teams_data.into_iter().group_by(|data| data.team_id.0);
+
+    let mut teams: Vec<Vec<crate::models::teams::TeamMember>> = vec![];
+
+    for (id, member_data) in &teams_groups {
+        let current_user =
+            get_user_from_headers(req.headers(), &**pool).await.ok();
```
Geometrically, line 131: why is this done every iteration?

src/routes/teams.rs
```diff
@@ -101,6 +101,65 @@ pub async fn team_members_get(
...
+
+        if let Some(user) = current_user {
+            let team_member = TeamMember::get_from_user_id(
+                crate::database::models::ids::TeamId(id),
+                user.id.into(),
+                &**pool,
+            )
+            .await
+            .map_err(ApiError::Database)?;
```
Geometrically, line 140: Is it possible to batch this with raw SQL- querying the database in a for loop is a bad idea and can be abused

Geometrically (changes requested):
This PR is querying the database really inefficiently.

It's not your fault- but avoid mass queries in for loops

src/routes/teams.rs
Ricky12Awesome: Can put that out of the loop

src/routes/teams.rs
Ricky12Awesome: Should be possible, working on that now

Geometrically (commented):
LGTM- does this hide member permissions (returns None for permissions)

Ricky12Awesome:
Yes, when I call the endpoint, it displays null for permissions, with `Authorization` set, it only shows for the teams the user is apart of

Geometrically (approved):

PR #376, opened 2022-06-18
Don't consider a user's name taken by self

John-Paul-R:
Relevant thread: https://discord.com/channels/734077874708938864/983131461647470592

Essentially, users cannot change the capitalization of their name, due to the current implementation of the "username taken" check.

---

Worth noting that I've not actually tested this locally due to local devenv issues I'm currently in the process of resolving.


Geometrically (changes requested):
Don't use an unwrap here. I know it won't panic, but it's pretty unintuitive. Just map the option to a bool and unwrap_or_default

Geometrically (approved):
lgtm! thanks for the PR

PR #380, opened 2022-06-23
Add Embedded dependency type

wafflecoffee:
I couldn't find any SQL tables or anything for dependency types, so I'm going to assume there aren't any


Geometrically (changes requested):
LGTM, but could you change pack autogen deps to be embedded too?

wafflecoffee:
I believe I changed it in the right places, if not let me know.

Geometrically (approved):
Thanks!

PR #382, opened 2022-06-24
Always return the username as fetched from the database

wafflecoffee:
Remake of #355 by @itsCryne as I tried to update the branch but somehow messed it up (blame Git)


Geometrically (approved):
lgtm! thanks for the pr

PR #392, opened 2022-07-06
Never return an empty 404 response

wafflecoffee:


Geometrically:
is this change needed? what's wrong with the current system?

wafflecoffee:
It was breaking Minotaur sometimes because of how Java deals with null stuff - if a project can't be found, it would throw a completely unrelated error because ErrorInfo was null.

It's also inconsistent with how nonexistent routes send a 404 with a body (example https://api.modrinth.com/v2/thisisafakeroute), but (for example) nonexistent projects don't have a body at all (example https://api.modrinth.com/v2/project/thisisafakeproject).

jamierocks:
Given that the documentation is clear that 404 responses don't have a response body, I don't particularly think this is neccasery either. I'm on the fence as to whether I would consider this a breaking change.

Geometrically:
Agreed, I know it's inconsistent with errors but this could break other tools. Maybe something for v3?

wafflecoffee:
Sounds good

jamierocks:
Standardising the error responses would be nice for v3! Don't forget the 400 Bad Requests :)

PR #393, opened 2022-07-06
Add featured loaders for use on frontend

wafflecoffee:


src/database/models/categories.rs
```diff
@@ -262,6 +263,7 @@ pub struct LoaderBuilder<'a> {
     pub name: Option<&'a str>,
     pub icon: Option<&'a str>,
     pub supported_project_types: Option<&'a [ProjectTypeId]>,
+    pub featured: &'a bool,
```
BasiqueEvangelist, line 266: Why are you using a reference here?

BasiqueEvangelist (commented):
Did you run `cargo sqlx prepare`?

src/database/models/categories.rs
wafflecoffee: Dunno, everything else was so I just copied it, removed it now

Geometrically (approved):

wafflecoffee:
Closing as hardcoding in other places (namely proxies) has made this obsolete/outdated/not inclusive of all of our use cases.

Prospector:
I'd suggest that this is still needed, headers would be a separate field I think (like categories have)

PR #394, opened 2022-07-08
Fix HTTP status code 200 → 204 on `/version/{id}/file`

venashial:
The status code should be 204 because no body is sent. This causes errors in Omorphia's API handling function.

Screen Shot 2022-07-07 at 9 09 45 PM


Geometrically (approved):
@Prospector Your turn to learn rust :smile:

PR #406, opened 2022-07-30
More project data

Geometrically:


migrations/20220725204351_more-project-data.sql
```diff
@@ -0,0 +1,33 @@
+-- Add migration script here
+ALTER TABLE mods_categories
+    ADD COLUMN is_additional BOOLEAN NOT NULL DEFAULT FALSE;
+
+ALTER TABLE mods
+    ADD COLUMN approved timestamptz NULL;
+
+UPDATE mods
+SET approved = published;
```
wafflecoffee, line 9:
```suggestion
    SET approved = published
    WHERE status = 'approved' OR status = 'unlisted';
```

migrations/20220725204351_more-project-data.sql
```diff
@@ -0,0 +1,37 @@
+-- Add migration script here
+ALTER TABLE mods_categories
+    ADD COLUMN is_additional BOOLEAN NOT NULL DEFAULT FALSE;
+
+ALTER TABLE mods
+    ADD COLUMN approved timestamptz NULL;
+
+ALTER TABLE categories
+    ADD COLUMN header varchar(256) NOT NULL DEFAULT 'Categories';
```
wafflecoffee, line 9: Maybe we could add a `description` field while we're at it?

src/search/mod.rs
```diff
@@ -74,6 +74,7 @@ pub struct UploadSearchProject {
     pub title: String,
     pub description: String,
     pub categories: Vec<String>,
+    pub display_categories: Vec<String>,
```
wafflecoffee, line 77: This seems really confusingly named. Instead of making secondary categories `categories` and primary categories `display_categories`, can't we just have `categories` and `secondary_categories`?

Geometrically: categories contains both additional and primary, so people can search by additional

migrations/20220725204351_more-project-data.sql
Geometrically: this pr is pretty big in scope, but this is out of scope, it's not really related to the changes needed for resourcepacks/plugins and we don't have descriptions for all the categories at the moment

PR #431, opened 2022-08-22
Persistent user settings

triphora:


wafflecoffee:
I have the best commit messages

Andre601:
> I have the best commit messages

Pffft... That's nothing.
image

Minenash:
I assume file mods will still be able to see the gh profile on the frontend?

src/database/models/team_item.rs
```diff
@@ -176,9 +178,17 @@ impl TeamMember {
...
+                            } else {
+                                None
+                            },
                             name: m.user_name,
-                            email: m.email,
+                            email: if m.public_email {
+                                m.email
+                            } else {
+                                None
+                            },
```
Geometrically, line 191: IMO a user shouldn't have an option to expose their email. I feel like the feature can be easily abused (to generate spam mailing lists). I think we should keep it private except for mods

src/models/settings.rs
```diff
@@ -0,0 +1,37 @@
+use serde::{Deserialize, Serialize};
```
Geometrically, line 1: isn't this file better fit for `users.rs`? it is user-only logic after all

src/routes/users.rs
```diff
@@ -544,3 +561,209 @@ pub async fn user_notifications(
...
+            user.id.0 as i64
+        )
+        .execute(&mut *transaction)
+        .await?;
+    }
+
+    transaction.commit().await?;
+
+    Ok(HttpResponse::NoContent().body(""))
+}
```
Geometrically, line 769: I don't see the use for these routes. The database already gets user settings with the user. Why not return settings when the user is requested in the first place by token & by ID (and just hide it if not authorized)?

Also- I would put the user settings inside of the user edit route. No need for a new route

Geometrically (changes requested):

Geometrically (approved):

migrations/20220822175311_user-settings.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE user_settings (
+    user_id bigint NOT NULL PRIMARY KEY UNIQUE,
+    public_github boolean NOT NULL DEFAULT TRUE,
+    theme varchar(32) NOT NULL DEFAULT 'system',
+    locale varchar(32) NOT NULL DEFAULT 'en-US'
```
Brawaru, line 5: Hey, in my Knossos i18n branch I have made automatic locale detection through `Accept-Language` header and `navigator.languages` browser API. It's also provided as one of the options in languages menu. In Nuxt plugin API it is not represented in any way, but you can change language to `'auto'` which internally switches to browser language / default locale (if detection failed) and turns on automatic mode, it then removes all stored language preferences since automatic detection is the default [^1].

Screenshot highlighting ‘Sync with browser’ option for languages

That being said, I'd assume it'd make sense for this default here to also be `'auto'` (special reserved value) or nullable instead, where null would represent absence of locale preference, thus making applications figure out which locale to use.
 
[^1]: [`changeLanguage` implementation](https://github.com/Brawaru/knossos/blob/feat/i18n-rewrite/modules/i18n/templates/plugin.main.js#L312-L336)

src/database/models/team_item.rs
jamierocks: Why not give the option? I rather enjoy responding to those that have taken the time to email me with a query on a project.

wafflecoffee:
I forgot about one critical thing that I won't be able to fix until tomorrow. This comment acts as a reminder to myself to fix that thing.

src/database/models/user_item.rs
```diff
@@ -68,16 +82,29 @@ impl User {
...
                     .unwrap_or_default(),
+                settings: if override_settings {
+                    Some(UserSettings {
+                        public_github: row.public_github,
+                        theme: FrontendTheme::from_str(&row.theme),
+                        locale: row.locale,
+                    })
+                } else {
+                    None
+                },
```
Geometrically, line 107: I'm not going to put this comment everywhere, But I think it's a lot better for code quality if these permissions checks happen in the actual conversion from a DBUser -> User. This stuff is repeated a lot and leaves the potential for when future calls are added for someone to forget putting the checks. If it's all in one place there's no problem

Geometrically (changes requested):

triphora:
Will revisit after auth

PR #434, opened 2022-08-25
Discord webhooks for new versions

triphora:


Geometrically (changes requested):
Webhook requests should be proxied and webhooks should be deleted on project/loader deletion

triphora:
Will revisit at a later date

PR #449, opened 2022-09-11
Move to SPDX licenses

### LINKED ISSUE #396: Support SPDX License Identifiers
### Is your feature request related to a problem? Please describe.

Licenses like 0BSD and dual-licensing under MIT OR Apache-2.0 aren't available on modrinth.

### Describe the solution you'd like

Honestly having an "SPDX" option, like "CUSTOM", would probably be the best. This lets modrinth have a curated list of licenses and "SPDX" can be used for anything extra.

### Describe alternatives you've considered

~~Putting "SPDX-License-Identifier: 0BSD" in "CUSTOM"... works... it technically sees it as an URL... it's... not ideal, tbh. :sweat_smile: ~~

the above stopped working when `javascript:` got patched out so it'd be really nice if we could have support for arbitrary `urn:` in license field.

### Additional context

Fun fact, did you know the copyright holder can make and distribute Minecraft mods under the GPL? After all, they're the copyright holder, so they're not actually bound by the license. Yes it's effectively ARR (but with permission to port to other games under GPL) but whatever. It shouldn't be up to modrinth to restrict these things.

triphora:
Found a way to do this without breaking API compat, so here it is, instead of waiting for v3
Even though it doesn't break API compat, we'll still want to do a frontend PR alongside this.

For search, as to not break compat with filters, only the first license is shown at all - e.g. `MIT OR Apache-2.0` gets shown as `MIT`.

Resolves MOD-129
Resolves #396


Geometrically (changes requested):
Add open source search filter- will leave code review later

migrations/20220910132835_spdx-licenses.sql
```diff
@@ -0,0 +1,40 @@
...
+UPDATE mods SET license_url = 'https://spdx.org/licenses/MIT.html' WHERE license_url = 'https://cdn.modrinth.com/licenses/mit.txt';
+UPDATE mods SET license_url = 'https://spdx.org/licenses/LGPL-3.0-only.html' WHERE license_url = 'https://cdn.modrinth.com/licenses/lgpl-3.txt';
+UPDATE mods SET license_url = 'https://spdx.org/licenses/LGPL-2.1-only.html' WHERE license_url = 'https://cdn.modrinth.com/licenses/lgpl-2.1.txt';
+UPDATE mods SET license_url = 'https://spdx.org/licenses/MPL-2.0.html' WHERE license_url = 'https://cdn.modrinth.com/licenses/mpl-2.txt';
+UPDATE mods SET license_url = 'https://spdx.org/licenses/ISC.html' WHERE license_url = 'https://cdn.modrinth.com/licenses/isc.txt';
+UPDATE mods SET license_url = 'https://spdx.org/licenses/Zlib.html' WHERE license_url = 'https://cdn.modrinth.com/licenses/zlib.txt';
+UPDATE mods SET license_url = 'https://spdx.org/licenses/GPL-2.0-only.html' WHERE license_url = 'https://cdn.modrinth.com/licenses/gpl-2.txt';
+UPDATE mods SET license_url = 'https://spdx.org/licenses/GPL-3.0-only.html' WHERE license_url = 'https://cdn.modrinth.com/licenses/gpl-3.txt';
```
Geometrically, line 35: The SPDX UI kinda sucks. I know it's not in the current UI, but since we're switching to a better system, can we add a route for getting the license text and just display it on the frontend in a popup? Then we can ditch license_url (or make it the api route for now) and have it on the frontend

src/models/projects.rs
```diff
@@ -218,7 +234,7 @@ impl SideType {
 #[derive(Serialize, Deserialize, Clone)]
 pub struct License {
     pub id: String,
-    pub name: String,
+    pub name: Option<String>,
```
Geometrically, line 237: Making this optional is an api breakage

src/search/indexing/local_import.rs
```diff
@@ -73,6 +72,16 @@ pub async fn index_local(
 
                     let project_id: crate::models::projects::ProjectId = ProjectId(m.id).into();
 
+                    let license = match m.license.split(" ").next() {
```
Geometrically, line 75: wouldn't this break if license ids had spaces in them? or is this not possible in a spdx license id

Geometrically: or, shouldn't this parse the entire spdx expression and see if it is osi approved? like if someone's code is lgpl but assets are arr, the filter shouldn't be open source

Geometrically (changes requested):

src/search/indexing/local_import.rs
triphora: This part is not used for getting the open source filter but for the license filter

migrations/20220910132835_spdx-licenses.sql
triphora: license_url is still useful for other reasons, imo, this could be changed in API v3 to have the license text instead though

src/search/indexing/local_import.rs
triphora: This just obtains the first license if there are spaces and the full thing otherwise. A space is what separates multiple licenses, but if there are no spaces it's just a single license.

src/routes/tags.rs
```diff
@@ -302,75 +301,41 @@ pub async fn game_version_delete(
...
+    if let Some(license) = spdx::license_id(&*license_id) {
+        return Ok(HttpResponse::Ok()
+            .content_type("text/plain")
+            .body(license.text()));
     }
+
+    Err(ApiError::InvalidInput(
+        "Invalid SPDX identifier specified".to_string(),
+    ))
 }
```
Geometrically, line 339: oops, I should've said in discord but for consistency this should probs be in JSON (just one field `body`)

migrations/20221126222222_spdx-licenses.sql
```diff
@@ -0,0 +1,27 @@
+ALTER TABLE mods ADD COLUMN license_new varchar(2048) DEFAULT 'LicenseRef-All-Rights-Reserved' NOT NULL;
+
+UPDATE mods SET license_new = licenses.short FROM licenses WHERE mods.license = licenses.id;
+
+UPDATE mods SET license_new = 'LicenseRef-Custom' WHERE license_new = 'custom';
+UPDATE mods SET license_new = 'LicenseRef-All-Rights-Reserved' WHERE license_new = 'arr';
```
Geometrically, line 6: On the frontend I think `LicenseRef` would look pretty ugly. Is there some way to map it without hardcoding?

Geometrically (changes requested):
Looks good, some  minor nitpicks

migrations/20221126222222_spdx-licenses.sql
triphora: Easy enough to do a substring find-and-delete, imo

Geometrically: got it, yeah makes sense. we can do something on the frontend for this

src/routes/tags.rs
```diff
@@ -302,75 +301,52 @@ pub async fn game_version_delete(
...
+        results.push(License {
+            short: short.to_string(),
+            name: name.to_string(),
+        });
+    }
 
-#[derive(serde::Deserialize)]
-pub struct LicenseData {
-    name: String,
+    HttpResponse::Ok().json(results)
```
Geometrically, line 321: in the future: let's remove this from the tags store/fetch in the frontend PR. Doesn't seem like it's needed anymore and this JSON is probs huge

Geometrically (approved):
Needs to be tested in staging, but looks good :)

PR #453, opened 2022-09-17
Statistics route

triphora:
Staging example:
example


Geometrically (changes requested):
join the queries with futures, no need for them to be executed sync

Geometrically (approved):

PR #502, opened 2022-12-18
Miscellaneous improvements and removals

triphora:
Got kind of carried away here when trying to do MOD-215, if there's anything you want me to revert don't hesitate to ask :sweat_smile:

Alternative commit title: Project Perses Part Πρώτος


src/routes/v1/versions.rs
```diff
@@ -242,44 +233,6 @@ pub async fn get_version_from_hash(
...
-
-    if let Some(id) = result {
-        Ok(HttpResponse::TemporaryRedirect()
-            .append_header(("Location", &*id.url))
-            .json(DownloadRedirect { url: id.url }))
-    } else {
-        Ok(HttpResponse::NotFound().body(""))
-    }
-}
-
```
Geometrically, line 282: this shouldn't be removed

Geometrically (changes requested):
lgtm besides one change

src/routes/v1/versions.rs
triphora: I thought that was a private route, but oki

triphora: Actually this is identical to the equivalent v2 route so I can keep it deleted and just call that

Geometrically: Not sure if it was private but iirc people did use it

src/routes/health.rs
```diff
@@ -1,31 +0,0 @@
...
-            "ready": false,
-            "reason": "Indexing is not finished"
-        });
-        return HttpResponse::InternalServerError().json(data);
-    }
-    HttpResponse::Ok().json(json!({
-        "ready": true,
-        "reason": "Everything is OK"
-    }))
-}
```
Geometrically, line 31: Also should keep this, it's useful in the future

Geometrically (changes requested):

Geometrically (approved):

PR #505, opened 2022-12-23
Improve error messages for report with invalid input

triphora:
While I was looking into MOD-184, I found that the database actually *does* check whether a report is using the right type! However, it gives a very unhelpful error message:
```json
{
  "error": "database_error",
  "description": "Database Error: error returned from database: insert or update on table \"reports\" violates foreign key constraint \"reports_mod_id_fkey\""
}
```

Thus, this PR makes it so that the error message is more friendly:
```json
{
  "error": "invalid_input",
  "description": "User could not be found: test"
}
```

I also merged the API v1 `POST /report` route into the v2 route so I didn't have to duplicate stuff and so that there wouldn't be conflicts with #502.

I'm not actually sure where the original issue with MOD-184 originates from. I wasn't able to reproduce the issue, which is even more confusing because I remember observing it happening quite often. It hasn't happened since rewrite-parity released, so I'm going to assume that the original issue is effectively fixed. In any event, this should prevent MOD-184 from happening again in case the database protection there messes up for whatever reason, and also makes the error message friendlier as an added bonus.


src/routes/reports.rs
```diff
@@ -64,32 +64,71 @@ pub async fn report_create(
...
-                serde_json::from_str::<ProjectId>(&format!(
-                    "\"{}\"",
-                    new_report.item_id
-                ))?
-                .into(),
+        ItemType::Project | ItemType::Mod => {
+            let project_id = serde_json::from_str::<ProjectId>(&format!(
+                "\"{}\"",
+                new_report.item_id
+            ))?;
```
Geometrically, line 71: I know this was in the code before, but we should just use `parse_base62` it's a lot better for all of these

src/routes/reports.rs
```diff
@@ -64,32 +64,71 @@ pub async fn report_create(
...
-            report.version_id = Some(
-                serde_json::from_str::<VersionId>(&format!(
-                    "\"{}\"",
-                    new_report.item_id
-                ))?
-                .into(),
+            let version_id = serde_json::from_str::<VersionId>(&format!(
+                "\"{}\"",
+                new_report.item_id
+            ))?;
```
Geometrically, line 93: see above

src/routes/reports.rs
```diff
@@ -64,32 +64,71 @@ pub async fn report_create(
...
-            report.user_id = Some(
-                serde_json::from_str::<UserId>(&format!(
-                    "\"{}\"",
-                    new_report.item_id
-                ))?
-                .into(),
+            let user_id = serde_json::from_str::<UserId>(&format!(
+                "\"{}\"",
+                new_report.item_id
+            ))?;
```
Geometrically, line 115: see above

src/routes/v1/mod.rs
```diff
@@ -122,7 +121,6 @@ pub fn moderation_config(cfg: &mut web::ServiceConfig) {
 }
 
 pub fn reports_config(cfg: &mut web::ServiceConfig) {
-    cfg.service(reports::reports);
-    cfg.service(reports::report_create);
+    cfg.service(super::reports::report_create);
     cfg.service(super::reports::delete_report);
```
Geometrically, line 125: couldn't we delete this one too? it's just a moderation route too right?

Geometrically (changes requested):
Looks good besides some minor changes

src/routes/reports.rs
```diff
@@ -64,32 +66,64 @@ pub async fn report_create(
...
-        ItemType::Project => {
-            report.project_id = Some(
-                serde_json::from_str::<ProjectId>(&format!(
-                    "\"{}\"",
-                    new_report.item_id
-                ))?
-                .into(),
+        ItemType::Project | ItemType::Mod => {
+            let project_id =
+                ProjectId(parse_base62(new_report.item_id.as_str())?);
```
Geometrically, line 71: Why even wrap it in `ProjectId`? I would just pass this straight to the DB call. All ids are identical (they are just wrappers of a u64). Same w/ everything below

Geometrically (changes requested):

src/routes/reports.rs
triphora: Addressed on Discord

Geometrically (approved):

PR #506, opened 2022-12-23
File types

Geometrically:
Fixes MOD-197

Primary files will eventually be moved to this system but need a little more work


src/routes/versions.rs
```diff
@@ -220,6 +220,8 @@ pub struct EditVersion {
     pub primary_file: Option<(String, String)>,
     pub downloads: Option<u32>,
     pub status: Option<VersionStatus>,
+    // algo, hash, status
+    pub file_types: Option<Vec<(String, String, Option<FileType>)>>,
```
triphora, line 224: This should be a separate struct instead of a tuple

triphora (changes requested):

triphora (approved):

PR #509, opened 2022-12-26
Fix query params giving plain text error

triphora:
Thanks to @masecla22 for reporting.


Geometrically (changes requested):
Looks good, could we also do the JSON config for those errors too? not sure if it's relevant now but would prevent errors like this not being wrapped in the correct struct

Geometrically (approved):

PR #511, opened 2022-12-28
Add notification for project status updates

BasiqueEvangelist:
Closes MOD-98

### DIFF
--- a/src/models/projects.rs
+++ b/src/models/projects.rs
@@ -317,11 +317,11 @@ impl ProjectStatus {
     }
     pub fn as_friendly_str(&self) -> &'static str {
         match self {
-            ProjectStatus::Approved => "Approved",
+            ProjectStatus::Approved => "Listed",
             ProjectStatus::Rejected => "Rejected",
             ProjectStatus::Draft => "Draft",
             ProjectStatus::Unlisted => "Unlisted",
-            ProjectStatus::Processing => "Processing",
+            ProjectStatus::Processing => "Under review",
             ProjectStatus::Unknown => "Unknown",
             ProjectStatus::Archived => "Archived",
             ProjectStatus::Withheld => "Withheld",
--- a/src/models/projects.rs
+++ b/src/models/projects.rs
@@ -315,6 +315,20 @@ impl ProjectStatus {
             ProjectStatus::Private => "private",
         }
     }
+    pub fn as_friendly_str(&self) -> &'static str {
+        match self {
+            ProjectStatus::Approved => "Approved",
+            ProjectStatus::Rejected => "Rejected",
+            ProjectStatus::Draft => "Draft",
+            ProjectStatus::Unlisted => "Unlisted",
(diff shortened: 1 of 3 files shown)


triphora:
Closes MOD-98

src/routes/projects.rs
```diff
@@ -551,6 +552,41 @@ pub async fn project_edit(
...
+                    ),
+                    link: format!(
+                        "/{}/{}",
+                        project_item.project_type,
+                        ProjectId::from(id)
+                    ),
+                    actions: vec![],
+                }
+                .insert_many(notified_members, &mut transaction)
+                .await?;
```
triphora, line 588: It only really needs to be sent if the status was set by a mod, imo. Currently, it gets sent for any status change, including for example draft -> processing

src/models/projects.rs
```diff
@@ -315,6 +315,20 @@ impl ProjectStatus {
             ProjectStatus::Private => "private",
         }
     }
+    pub fn as_friendly_str(&self) -> &'static str {
+        match self {
+            ProjectStatus::Approved => "Approved",
+            ProjectStatus::Rejected => "Rejected",
+            ProjectStatus::Draft => "Draft",
+            ProjectStatus::Unlisted => "Unlisted",
+            ProjectStatus::Processing => "Processing",
```
triphora, line 324:
```suggestion
            ProjectStatus::Approved => "Listed",
            ProjectStatus::Rejected => "Rejected",
            ProjectStatus::Draft => "Draft",
            ProjectStatus::Unlisted => "Unlisted",
            ProjectStatus::Processing => "Under review",
```

BasiqueEvangelist: Oh right, they're not called that...

src/routes/projects.rs
BasiqueEvangelist: I've made it so that it only sends notifications if the user changing the status isn't in the team

triphora (approved):

src/routes/projects.rs
```diff
@@ -551,6 +552,42 @@ pub async fn project_edit(
                     }
                 }
 
+                if team_member.map(|x| !x.accepted).unwrap_or(true) {
+                    let user_id: database::models::ids::UserId = user.id.into();
+                    let notified_members = sqlx::query!(
+                        "
+                        SELECT tm.user_id id
+                        FROM team_members tm
+                        WHERE tm.team_id = $1 AND tm.user_id <> $2 AND tm.accepted
```
Geometrically, line 561: Could we use `!=` instead of `<>` here? it's more readable and consistent w/ the codebase. Also, does the check even need to be there in the first place? Don't see a case in this code path where it would send a notif to the team member who set it (since it only sends notifs if team members haven't sent it)

Geometrically (approved):

Geometrically (changes requested):
Meant to request changes, see comment above

Geometrically (approved):

PR #517, opened 2023-01-05
Bulk Editing + Random Projects Route

Geometrically:
Fixes MOD-234

### DIFF
--- a/src/routes/projects.rs
+++ b/src/routes/projects.rs
@@ -2,7 +2,6 @@ use crate::database;
 use crate::database::models::notification_item::NotificationBuilder;
 use crate::file_hosting::FileHost;
 use crate::models;
-use crate::models::ids::UserId;
 use crate::models::projects::{
     DonationLink, Project, ProjectId, ProjectStatus, SearchRequest, SideType,
 };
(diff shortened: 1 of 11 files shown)


src/routes/projects.rs
```diff
@@ -1085,102 +1113,432 @@ pub async fn project_edit(
...
+    #[serde(
+        default,
+        skip_serializing_if = "Option::is_none",
+        with = "::serde_with::rust::double_option"
+    )]
+    #[validate(
+        custom(function = "crate::util::validate::validate_url"),
+        length(max = 2048)
+    )]
+    pub license_url: Option<Option<String>>,
```
triphora, line 1188: License URL shouldn't be bulk editable, authors shouldn't be able to blindly set a license URL without having right next to it the license they have set

src/routes/projects.rs
```diff
@@ -1085,102 +1113,432 @@ pub async fn project_edit(
...
+            .execute(&mut *transaction)
+            .await?;
+        }
+
+        if let Some(wiki_url) = &bulk_edit_project.wiki_url {
+            sqlx::query!(
+                "
+                    UPDATE mods
+                    SET wiki_url = $1
+                    WHERE (id = $2)
```
triphora, line 1501: Formatting is a little janky for each of these but nbd

triphora (changes requested):

src/routes/projects.rs
Geometrically: oops, true

PR #522, opened 2023-01-08
Add project color to embed accent color

### LINKED ISSUE #521: Add a category-based color for discord web-hook embeds
### Is your suggested enhancement related to a problem? Please describe.

Official discord server of Modrinth has a channel `#new-projects`, with embeds of every new project that has been verified. All embeds use modrinth's accent color (green), making it hard to distinguish between different types of projects (mods, shaders, etc.). The only visual information is the embed's footer.

### Describe the solution you'd like

Add a unique color for every category of submission, so that when you are exploring the channel you can easily understand what kind of project you are looking at.

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

triphora:


Prospector:
Hmm, this isn't exactly what the suggestion was. They wanted the color to represent the project type so you could see what kind of project it is more easily at a glance.

triphora:
I do agree that it's not what was being asked for, but I don't know if colors for project types is the best way of going about it.

Geometrically (approved):

MagnusHJensen:
It looks good.
But again it does not make sense, since there currently is no way of setting this color per project.
So maybe make it the project color or map it to a color based on project type.
`project.color.unwrap_or(get_color_from_project_type(project.type))`

triphora:
The project color is auto generated from the icon

MagnusHJensen:
> The project color is auto generated form the icon

Hmm, fair enough.
I still don't think it makes sense for the embeds labrinth sends out in discord. I think it should only be split by project type.

PR #526, opened 2023-01-18
Allow mods to edit publishing dates of versions

Geometrically:


Prospector:
I don't think publishing dates should change, they should remain the date that they were approved by moderators/appeared on Modrinth.

Creation dates should be the ones that can be retroactively edited to match their original creation/publication

PR #527, opened 2023-01-25
add atom feed endpoint

BasiqueEvangelist:
currently in draft since I'm unsure whether I should use a markdown parser for the content or not
\+ there are some todo lines in

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -365,6 +365,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "atom_syndication"
+version = "0.12.0"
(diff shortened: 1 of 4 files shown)


BasiqueEvangelist:
How do you run that check locally :thonk:
anyways, that's even more reasons for it to be in draft

triphora:
I think it would just be better for this to be in the frontend, tbh. https://github.com/jpmonette/feed is quite great and doesn't require anywhere near as much maintenance burden.

BasiqueEvangelist:
Closing, as the Knossos PR will most likely replace this

PR #535, opened 2023-02-10
Validate minimum size for icon image

MagnusHJensen:
Should a maximum size also be set?
Should it always enforce 1:1 ratios?
PolyMC mentioned they saw a 1920x1920 image once. (Lots of unnecessary data to store)

This PR adds the minimum icon size validation to the icon patch request.
Alongside it, it also adds a small image utility function that extracts the width, and height from any image.

_Note:
Next step would need to be a PR in knossos, that removes the image if the icon update failed.
Not sure I will personally do that, however I will open an issue once this is merged to track it._

**Media:**

https://user-images.githubusercontent.com/32039051/218141975-dadbe5b3-131b-43aa-87ff-ff33e4a4e69c.mov

Closes: #216

### DIFF
--- a/src/routes/projects.rs
+++ b/src/routes/projects.rs
@@ -1655,6 +1655,15 @@ pub async fn project_icon_edit(
         )
         .await?;
 
+        let (image_width, image_height) =
+            crate::util::img::get_image_size(&bytes)?;
+
+        if image_width < 100 || image_height < 100 {
+            return Err(ApiError::InvalidInput(
+                "Icon must meet the minimum 100x100px requirements".to_string(),
+            ));
+        }
+
         let color = crate::util::img::get_color_from_img(&bytes)?;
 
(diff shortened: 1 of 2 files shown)


triphora:
Sorry, but I'm going to have to reject this PR. Authors oftentimes upload pixel art or one of their textures as the project icon. Forcing 100x100 is not very reasonable—rather, if it displays poorly on knossos, the CSS should be changed there to fix that.

Thanks for the contribution nonetheless.

Regarding the questions asked in the OP:
> Should a maximum size also be set?

The 256KiB maximum size is fine for this.

> Should it always enforce 1:1 ratios?

I don't think this is needed? As shown in the video of the OP, even icons which are not 1:1 display well enough.

> PolyMC mentioned they saw a 1920x1920 image once. (Lots of unnecessary data to store)

Still always under 256KiB.

PR #536, opened 2023-02-12
Allow for retrieving multiple users by usernames

MagnusHJensen:
**Description:**
This PR introduces behaviour found in the singular user get request, were an ID or username is allowed, into the multiple `GET users?ids=[...]` route.

**Media:**

Getting multiple users solely on IDs still work:
image

Getting multiple users solely on usernames now work:
image

**If mixing IDs and usernames, and the username is less than 11 characters, the id's take precedence.**
image

**If mixing IDs and usernames, and the username is 11 characters or longer (not base62 decodable - overflow) the usernames take precedence.**
image

**Follow-up:**
- Create a PR in [modrinth/docs](https://github.com/modrinth/docs) to update the OpenAPI.yaml spec.
Specifically updating the spec to this:
```yaml
/users:
    get:
      summary: Get multiple users
      operationId: getUsers
      tags:
        - users
      parameters:
        - in: query
          name: ids
          description: |
            The IDs or Usernames of the users
            If the array is mixed with id's and usernames, id's will take precedence, unless the username is less than 10 characters.
          schema:
            type: array
            items:
              type: string
            example: "[\"AABBCCDD\", \"EEFFGGHH\"]"
          required: true
      responses:
        "200":
          description: Expected response to a valid request
          content:
            application/json:
              schema:
                type: array
                items: *User
```

### DIFF
--- a/src/routes/users.rs
+++ b/src/routes/users.rs
@@ -41,16 +41,39 @@ pub async fn users_get(
     web::Query(ids): web::Query<UserIds>,
     pool: web::Data<PgPool>,
 ) -> Result<HttpResponse, ApiError> {
-    let user_ids = serde_json::from_str::<Vec<UserId>>(&ids.ids)?
-        .into_iter()
-        .map(|x| x.into())
-        .collect();
+    let mut users_data;
 
-    let users_data = User::get_many(user_ids, &**pool).await?;
+    let decoded_user_ids = serde_json::from_str::<Vec<UserId>>(&ids.ids);
+    if decoded_user_ids.is_ok() {
+        let user_ids = decoded_user_ids
+            .unwrap()
+            .into_iter()
+            .map(|x| x.into())
+            .collect();
+        users_data = Some(User::get_many(user_ids, &**pool).await?);
+
+        if let Some(users) = &users_data {
+            if users.is_empty() {
+                let user_names = serde_json::from_str::<Vec<String>>(&ids.ids)?
+                    .into_iter()
+                    .map(|x| x.to_lowercase())
+                    .collect();
+                users_data = Some(
+                    User::get_many_from_usernames(user_names, &**pool).await?,
+                );
+            }
+        }
+    } else {
+        let user_names = serde_json::from_str::<Vec<String>>(&ids.ids)?
+            .into_iter()
+            .map(|x| x.to_lowercase())
+            .collect();
+        users_data =
+            Some(User::get_many_from_usernames(user_names, &**pool).await?);
+    }
 
     let users: Vec<crate::models::users::User> =
-        users_data.into_iter().map(From::from).collect();
-
+        users_data.unwrap().into_iter().map(From::from).collect();
     Ok(HttpResponse::Ok().json(users))
 }
 
--- a/sqlx-data.json
+++ b/sqlx-data.json
@@ -7747,6 +7747,110 @@
     },
     "query": "SELECT EXISTS(SELECT 1 FROM notifications WHERE id=$1)"
   },
+  "fc53c5521be2979443c6c3a80b6f798b6612a41b7ea80d573d3a8e7a37e1dc54": {
+    "describe": {
+      "columns": [
+        {
+          "name": "id",
+          "ordinal": 0,
+          "type_info": "Int8"
+        },
+        {
+          "name": "github_id",
+          "ordinal": 1,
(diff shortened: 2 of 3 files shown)


src/routes/users.rs
```diff
@@ -41,16 +41,39 @@ pub async fn users_get(
...
 ) -> Result<HttpResponse, ApiError> {
-    let user_ids = serde_json::from_str::<Vec<UserId>>(&ids.ids)?
-        .into_iter()
-        .map(|x| x.into())
-        .collect();
-
-    let users_data = User::get_many(user_ids, &**pool).await?;
+    let mut users_data;
+
+    let decoded_user_ids = serde_json::from_str::<Vec<UserId>>(&ids.ids);
```
MagnusHJensen, line 46: due to usernames often being longer than 10 characters, this will throw an error.
And we therefore search on usernames instead, not sure if (also see media) that makes the most sense? It would of course be explained in the API Docs, could be I'm missing some fancy way of doing this, that could not lead to the same slightly "unexpected" behaviour.

src/routes/users.rs
```diff
@@ -41,16 +41,39 @@ pub async fn users_get(
...
+    if decoded_user_ids.is_ok() {
+        let user_ids = decoded_user_ids
+            .unwrap()
+            .into_iter()
+            .map(|x| x.into())
+            .collect();
+        users_data = Some(User::get_many(user_ids, &**pool).await?);
+
+        if let Some(users) = &users_data {
+            if users.is_empty() {
```
MagnusHJensen, line 56: This could benefit from let chaining, but that is not available yet.

src/routes/users.rs
triphora: That seems like a really poor, bug-prone way of doing it. Isn't there a get_project_from_id_or_slug method (or something similar) for projects? How is that implemented?

triphora (changes requested):

src/routes/users.rs
MagnusHJensen: There is not.
This is similar to how fetching a single user is implemented. 
```rust
    let string = info.into_inner().0;
    let id_option: Option<UserId> =
        serde_json::from_str(&format!("\"{string}\"")).ok();

    let mut user_data;

    if let Some(id) = id_option {
        user_data = User::get(id.into(), &**pool).await?;

        if user_data.is_none() {
            user_data = User::get_from_username(string, &**pool).await?;
        }
    } else {
        user_data = User::get_from_username(string, &**pool).await?;
    }

    if let Some(data) = user_data {
        let response: crate::models::users::User = data.into();
        Ok(HttpResponse::Ok().json(response))
    } else {
        Ok(HttpResponse::NotFound().body(""))
    }
```

Geometrically: That isn't what the code actually does. It checks if it is parable as an ID (as ids always take precedence over a slug). If a project isn't found by ID, it tries to find it using a slug.

Geometrically (changes requested):
Don't really like the approach here and it is buggy

src/routes/users.rs
MagnusHJensen: It it though.

The single user get
Tries to first parse the string into a UserId. (This would throw an error if the value is greater than 10 characters), since we directly calling `.ok()` the error is discarded, but a None value is then provided.

If it matched an id (less than 11), meaning the string contains a value (which technically could be a small username), it tries to first find a user from id, if that fails (due to low character username), then it fetches on username.

If originally the value was more than 10 characters, (often a username) it would go straight to finding a user based on their username.

Which is identical to what I'm doing here for multiple users.
I first try and parse the list of values into a list of user_ids, if that fails (errors, since I'm not calling `.ok()` directly) then I go straight to fetching users on their usernames.

if it doesn't fail originally, it would automatically go into and try and fetch users based on their ids. (At this point that array could be a mix of id's and low character usernames)
If the list contains ANY id's, the get_many call will find at least one, and exit.
If the list ONLY contains low character names, then the get_many (based on id's) will not fetch any, trying to fetch users based on usernames.

I'm going to say that logic is identical, between the single user and multiple users. (Except for the case where the array is mixed).

_On a completely side note, I would say this logic should be abstracted away to a method inside the User model, since the "controller/router" does not need to care about this. it should just know if any users is retrieved and then return them._

Geometrically:
superseded by #649

PR #537, opened 2023-02-12
Fix version name can be empty string

MagnusHJensen:
**Description:**
Validates the version name, to check if trimming the name results in the string being empty.
Also trims the name on insert, to remove whitespace before and after name.

**Media:**

https://user-images.githubusercontent.com/32039051/218306547-d29a6ef7-9211-401c-bc9c-8e4c83617a1f.mov

Closes: #169

### DIFF
--- a/src/routes/version_creation.rs
+++ b/src/routes/version_creation.rs
@@ -41,7 +41,10 @@ pub struct InitialVersionData {
         regex = "crate::util::validate::RE_URL_SAFE"
     )]
     pub version_number: String,
-    #[validate(length(min = 1, max = 64))]
+    #[validate(
+        length(min = 1, max = 64),
+        custom(function = "crate::util::validate::validate_name")
+    )]
     #[serde(alias = "name")]
     pub version_title: String,
     #[validate(length(max = 65536))]
@@ -275,7 +278,7 @@ async fn version_create_inner(
                         file_name: None,
                     })
                     .collect::<Vec<_>>();
-
+                println!("{}", version_create_data.version_title.clone());
                 version_builder = Some(VersionBuilder {
                     version_id: version_id.into(),
                     project_id,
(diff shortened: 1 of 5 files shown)


MagnusHJensen:
The same validation should now run on project/version creation, and project/version editing.

src/routes/version_creation.rs
```diff
@@ -275,7 +278,7 @@ async fn version_create_inner(
                         file_name: None,
                     })
                     .collect::<Vec<_>>();
-
+                println!("{}", version_create_data.version_title.clone());
```
triphora, line 281: Debug line left in
```suggestion

```

triphora (changes requested):

src/routes/version_creation.rs
MagnusHJensen: Fixed

triphora (approved):

Geometrically (changes requested):
It looks like this only applies to version editing, not creation

MagnusHJensen:
It also applies to version creation, and assuming project creation due to no validation/trimming.
Just tested it creating a version with blank name with spaces, and it was allowed.

Geometrically (approved):

src/util/validate.rs
```diff
@@ -97,3 +97,30 @@ pub fn validate_url(value: &str) -> Result<(), validator::ValidationError> {
 
     Ok(())
 }
+
+pub fn validate_name(value: &str) -> Result<(), validator::ValidationError> {
+    if value.trim().is_empty() {
+        return Err(validator::ValidationError::new(
+            "Name can not contain only whitespace.",
```
triphora, line 104:
```suggestion
            "Name cannot contain only whitespace.",
```

PR #539, opened 2023-02-12
feature(version): include project_name in version response

MagnusHJensen:
**Description:**
Adds a new field to the version struct, `project_name` that holds the name of the project the version is attached to.
This is to avoid extra queries, for fetching projects based on ID's just to get their names.

**Media:**

image

Closes: #416

### DIFF
--- a/src/routes/version_creation.rs
+++ b/src/routes/version_creation.rs
@@ -276,9 +276,18 @@ async fn version_create_inner(
                     })
                     .collect::<Vec<_>>();
 
+                let project_name = sqlx::query!(
+                    "SELECT m.title FROM mods m WHERE m.id = $1",
+                    project_id as models::ProjectId
+                )
+                .fetch_one(&mut *transaction)
+                .await?
+                .title;
+
                 version_builder = Some(VersionBuilder {
                     version_id: version_id.into(),
                     project_id,
+                    project_name,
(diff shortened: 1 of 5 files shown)


src/routes/version_creation.rs
```diff
@@ -276,9 +276,18 @@ async fn version_create_inner(
                     })
                     .collect::<Vec<_>>();
 
+                let project_name = sqlx::query!(
+                    "SELECT m.title FROM mods m WHERE m.id = $1",
+                    project_id as models::ProjectId
+                )
+                .fetch_one(&mut *transaction)
+                .await?
+                .title;
```
MagnusHJensen, line 285: Maybe this can be combined with the exists query on lines [185-190]

Geometrically:
Closing because this isn't how our API is structured. If someone wants the project title, they should make another request. Thanks for the PR though. In the future- I would recommend talking through a feature on Discord before implementing it.

MagnusHJensen:
My bad, I just thought since the issue had been open for 6 months, it was up for grabs.
Even labelled.

Geometrically:
It's all good :) Users can self-label. We should go out and clear out issues more that we won't be implementing, we do it a lot for knossos but we should do it more for labrinth

PR #541, opened 2023-02-15
Add API V1 flickers

Geometrically:


src/routes/v1/mod.rs
```diff
@@ -9,6 +11,23 @@ mod versions;
 pub fn v1_config(cfg: &mut web::ServiceConfig) {
     cfg.service(
         web::scope("api/v1")
+            .wrap_fn(|req, srv| {
+                let current_minute = Utc::now().timestamp() / 60;
```
triphora, line 15: I'm pretty sure there's a .minutes method you can call directly

src/routes/v1/mod.rs
```diff
@@ -9,6 +11,23 @@ mod versions;
...
+
+                if current_minute % 10 > 5 {
```
triphora, line 17: Hmm, this seems too frequent. I was more thinking 5-10 minutes per hour, and then later 6-12 hours at a time. Do you think this approach is better, though?

src/routes/v1/mod.rs
```diff
@@ -9,6 +11,23 @@ mod versions;
...
+                    srv.call(req).boxed_local()
+                } else {
+                    async {
+                        Ok(
+                            req.into_response(
+                                HttpResponse::Gone()
+                                    .content_type("application/json")
+                                    .body(r#"{"error":"api_deprecated","description":"API V1 was deprecated several months ago. Please upgrade to V2. Check out https://docs.modrinth.com/docs/migrations/v1-to-v2/"}"#)
```
triphora, line 25: IMO this error is a bit too user-unfriendly. It's less a matter of developers still using it and more a matter of people using outdated applications which use it. Might I suggest the following?
```suggestion
                                    .body(r#"{"error":"api_deprecated","description":"You are using an application that uses an outdated version of Modrinth's API. Please either update it or switch to another application. For developers: https://docs.modrinth.com/docs/migrations/v1-to-v2/"}"#)
```

triphora (changes requested):

src/routes/v1/mod.rs
Geometrically: Doesn't exist

src/routes/v1/mod.rs
Geometrically: I did this bc it was easiest to code, but if you want you can commit to this branch and change it. I think either works but more frequent is probably better IMO (most applications using API V1 would already have seen the messages appended to every data field)

src/routes/v1/mod.rs
Geometrically: thanks! yeah I agree

src/routes/v1/mod.rs
triphora: https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.minute

src/routes/v1/mod.rs
triphora: Yeah that's fair

src/routes/v1/mod.rs
triphora: Doesn't matter too much, if it works it works, this is temporary code anyway

triphora (approved):

src/routes/v1/mod.rs
MagnusHJensen: `Returns the minute number from 0 to 59.`

triphora:
```suggestion
                let current_minute = Utc::now().minute();
```

src/routes/v1/mod.rs
```diff
@@ -1,4 +1,6 @@
-use actix_web::web;
+use actix_web::{dev::Service, web, HttpResponse};
+use chrono::Utc;
```
triphora, line 2:
```suggestion
use chrono::{Timelike, Utc};
```

PR #551, opened 2023-02-23
Add links on user object

MagnusHJensen:
**Description:**
This is currently very simple in the sense, it only adds these fields then allows for updating them.

This is a stepping stone towards completing the feature in #278
I'm doing this in multiple PR's, to reduce code review amount, and also for faster feedback.

Prospector mentioned something about Project templates, instead of user links? That would invalidate this PR if that is the direction wanted.

Logic for getting either this or project, would be a follow up PR.

~~Does not close:: #278~~


MagnusHJensen:
Ready for review, see updated description.

Geometrically:
feature will likely not be added to modrinth, closing for now. feel free to discuss in discord

PR #556, opened 2023-03-03
Potentially fix total_hits number in search

triphora:


triphora:
Nope, doesn't work. -> https://github.com/meilisearch/meilisearch/discussions/3559

PR #558, opened 2023-03-04
Perses finale

triphora:
Removal of v1, moving of v2, and maybe the start of v3?

Still very WIP and still trying to work out what the scope of this PR is. Making this PR because I much prefer the GitHub diff view over using plain `git diff`.

Closes MOD-217
Closes MOD-221
Closes MOD-251


src/routes/mod.rs
```diff
@@ -1,200 +1,35 @@
-mod v1;
-pub use v1::v1_config;
+use crate::file_hosting::FileHostingError;
+use actix_web::http::{Method, StatusCode};
```
Geometrically, line 2: unused import `Method`

src/routes/mod.rs
```diff
@@ -1,200 +1,35 @@
...
+use actix_web::{web, HttpResponse};
+use chrono::{Timelike, Utc};
```
Geometrically, line 4: unused imports

Geometrically (changes requested):
looks good besides some minor things. have you tested that the new route registration works?

triphora:
As far as I can tell it works

Geometrically (approved):

PR #562, opened 2023-03-07
Fixed an issue with colliding slugs when modifying a project

### LINKED ISSUE #561: Project slug collides issue on modification request
### Describe the bug

The collision check on projects will error out if you send a modify request with the slug of the project you are changing!

### Steps to reproduce

1. Make project (slug: example)
2. Send modify request to project containing slug example in it
3. Errors out

### Expected behavior

The slug of the project remains unchanged

### Additional context

_No response_

masecla22:
Fixes #561. 

Covered by Modrinth4J test cases at 
https://github.com/masecla22/Modrinth4J/blob/master/src/test/java/masecla/modrinth4j/client/ProjectEndpointsTests.java#L302-L316

### DIFF
--- a/src/routes/projects.rs
+++ b/src/routes/projects.rs
@@ -893,11 +893,13 @@ pub async fn project_edit(
                 }
 
                 {
+                    // We exclude the ID of the project we are editing, so that we can edit the slug to the same value
                     let results = sqlx::query!(
-                        "
-                      SELECT EXISTS(SELECT 1 FROM mods WHERE slug = LOWER($1))
+                      "
+                      SELECT EXISTS(SELECT 1 FROM mods WHERE slug = LOWER($1) AND id != $2)
                       ",
-                        slug
+                        slug,
+                        id as database::models::ids::ProjectId,
                     )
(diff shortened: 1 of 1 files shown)


src/routes/projects.rs
```diff
@@ -893,11 +893,13 @@ pub async fn project_edit(
                 }
 
                 {
```
triphora, line 895: Wouldn't it be better to do the check here in code instead of in the query? We can bypass this entirely if the new slug is the same as the old slug.

masecla22: It may just be my unfamiliarity with the code, though I didn't see anything in that scope that would allow me to check the _current_ slug of the project without an additional query.

triphora: `project_item` is the current project

triphora (approved):

triphora (approved):

triphora:
..why did I rereview that?

Geometrically (approved):

masecla22:
> ..why did I rereview that?

No clue

PR #567, opened 2023-03-25
Add report + moderation messaging

Geometrically:
Fixes MOD-117
Fixes MOD-156
Fixes MOD-157


src/models/threads.rs
```diff
@@ -0,0 +1,75 @@
...
+        }
+    }
+
+    pub fn from_str(string: &str) -> ThreadType {
+        match string {
+            "required" => ThreadType::Report,
+            "optional" => ThreadType::Project,
+            "unsupported" => ThreadType::DirectMessage,
+            _ => ThreadType::DirectMessage,
+        }
```
triphora, line 73: These need to be updated

src/routes/v2/mod.rs
```diff
@@ -32,6 +33,7 @@ pub fn config(cfg: &mut actix_web::web::ServiceConfig) {
             .configure(teams::config)
             .configure(users::config)
             .configure(version_file::config)
-            .configure(versions::config),
+            .configure(versions::config)
+            .configure(threads::config),
```
triphora, line 37: Keep the alphabetical order :irritatered:

src/routes/v2/threads.rs
```diff
@@ -0,0 +1,200 @@
...
+    }
+    Ok(HttpResponse::NotFound().body(""))
+}
+
+#[derive(Deserialize)]
+pub struct NewThreadMessage {
+    pub body: MessageBody,
+}
+
+#[patch("{id}")]
```
triphora, line 126: Seems like this should be a POST route

triphora (dismissed):
Is there be a way for moderators to see existing *project* threads (not report threads) that are unresolved? Also, how does opening/closing project threads work?

Prospector (changes requested):
Thread messages need a way to differentiate anonymous moderator from deleted user or a user that is null for potentially other reasons.

Also, it'd be nice if the project status change message type included the old project status in the message

thesuzerain (approved):

PR #573, opened 2023-04-15
Overhaul notifs + threads fixes

Geometrically:


src/models/threads.rs
```diff
@@ -45,6 +49,7 @@ pub enum MessageBody {
         old_status: ProjectStatus,
     },
     ThreadClosure,
+    ThreadReOpen,
```
Prospector, line 52: Reopen is more commonly one word (and thread_re_open is weird in snake_case)

Prospector (changes requested):

Prospector (approved):

PR #575, opened 2023-04-18
Return pending TMs, fix notifs serde

Geometrically:


src/routes/v2/teams.rs
```diff
@@ -108,7 +108,10 @@ pub async fn team_members_get(
 
     let team_members: Vec<_> = members_data
         .into_iter()
-        .filter(|x| x.accepted)
+        .filter(|x| {
+            x.accepted
+                || current_user.map(|y| y.id == x.user.id).unwrap_or(false)
```
triphora, line 113: Clippy sure doesn't like this

triphora (changes requested):

triphora (approved):

PR #582, opened 2023-04-21
Fix edit bug staging

Geometrically:


src/routes/v2/projects.rs
```diff
@@ -1073,7 +1076,10 @@ pub async fn project_edit(
             }
 
             if let Some(moderation_message_body) = &new_project.moderation_message_body {
-                if !user.role.is_mod() && project_item.inner.status != ProjectStatus::Approved {
+                if !user.role.is_mod()
+                    && (project_item.inner.status != ProjectStatus::Approved
```
triphora, line 1080: Could this be changed to `is_approved`? Unrelated bug but would be easy to knock out here

PR #587, opened 2023-05-06
Insert release versions as major by default

triphora:
Resolves MOD-321
Makes new release versions major by default. If something is inserted as major when we don't want it to be, this can be reverted in the DB later.


Geometrically (changes requested):
I don't think this will work, as it will edit every version and set it to major. Has it been tested with existing data?

triphora:
You're right, this is also kinda pointless

PR #622, opened 2023-05-21
Loader + game version combo search field

stairman06:
Required backend implementation for https://github.com/modrinth/knossos/issues/447.

My SQL might be poor so feel free to critique.

### DIFF
--- a/sqlx-data.json
+++ b/sqlx-data.json
@@ -590,178 +590,6 @@
     },
     "query": "\n        SELECT EXISTS(SELECT 1 FROM mod_follows mf WHERE mf.follower_id = $1 AND mf.mod_id = $2)\n        "
   },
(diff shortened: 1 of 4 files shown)


Geometrically (changes requested):
why not just use the existing joins for this?

stairman06:
I don't believe this is possible with the existing joins? They don't have the right association between version and game version/loader. But I'll need to check and look at it again.

Geometrically:
couldn't you only join the combos and then extract the rest of the versions and loaders in the rust side of code?

stairman06:
I think with that we'd end up with the loader ID and the game version ID, which aren't what we want. To actually convert those to the loader name and game version name we have to ask the database AFAIK. But maybe I'm misunderstanding?

Geometrically:
Superseded by #731

PR #633, opened 2023-06-10
add created_timestamp to facets

kevoh1516:
I added the `created_timestamp` attribute to the facets so people can choose to only search for projects published within the last `x` amount of time


triphora:
Supercede by #708

PR #647, opened 2023-07-06
Initial Auth Impl + More Caching

Geometrically:


src/auth/session.rs
```diff
@@ -0,0 +1,80 @@
...
+        .headers()
+        .get("cf-ipcountry")
+        .and_then(|x| x.to_str().ok());
+    let city = req.headers().get("cf-ipcity").and_then(|x| x.to_str().ok());
+
+    let user_agent = req
+        .headers()
+        .get("user-agent")
+        .and_then(|x| x.to_str().ok())
+        .ok_or_else(|| AuthenticationError::InvalidCredentials)?;
```
thesuzerain, line 38: Could we make this a separate err? If I were trying to log in and I forgot the user agent, invalid credentials being the error would be confusing and make me think I had the wrong password or mistyped my token or smthn.

If we dont anticipate this being used by the API ever directly then its fine

src/auth/validate.rs
```diff
@@ -0,0 +1,108 @@
...
+        &redis,
+    )
+    .await?
+    .ok_or_else(|| AuthenticationError::InvalidCredentials)?;
+
+    let user = User {
+        id: UserId::from(db_user.id),
+        github_id: db_user.github_id.map(|x| x as u64),
+        // discord_id: minos_user.discord_id,
+        // google_id: minos_user.google_id,
```
thesuzerain, line 35: We can delete these now

src/auth/validate.rs
```diff
@@ -0,0 +1,108 @@
...
+    executor: E,
+    redis: &deadpool_redis::Pool,
+) -> Result<User, AuthenticationError>
+where
+    E: sqlx::Executor<'a, Database = sqlx::Postgres> + Copy,
+{
+    let token: Option<&HeaderValue> = headers.get(AUTHORIZATION);
+
+    // Fetch DB user record and minos user from headers
+    let db_user = get_user_record_from_bearer_token(
```
thesuzerain, line 20: Cookie implementation is still TODO, right? Using the bearer token as the primary authentication point is totally good but we should definitely have a way to get the mra_ bearer token from a cookie

src/database/models/session_item.rs
```diff
@@ -0,0 +1,310 @@
...
+
+    pub async fn get_many<'a, E, T: ToString>(
+        session_strings: &[T],
+        exec: E,
+        redis: &deadpool_redis::Pool,
+    ) -> Result<Vec<Session>, DatabaseError>
+    where
+        E: sqlx::Executor<'a, Database = sqlx::Postgres>,
+    {
+        use futures::TryStreamExt;
```
thesuzerain, line 132: clippy error here, unused use futures::TryStreamExt;

src/database/models/session_item.rs
```diff
@@ -0,0 +1,310 @@
...
+        let res = cmd("GET")
+            .arg(format!("{}:{}", SESSIONS_USERS_NAMESPACE, user_id.0))
+            .query_async::<_, Option<Vec<i64>>>(&mut redis)
+            .await?;
+
+        if let Some(res) = res {
+            return Ok(res.into_iter().map(|x| SessionId(x)).collect());
+        }
+
+        use futures::TryStreamExt;
```
thesuzerain, line 268: clippy

src/models/users.rs
```diff
@@ -48,11 +47,11 @@ pub struct User {
...
     pub payout_data: Option<UserPayoutData>,
     pub github_id: Option<u64>,
-    pub discord_id: Option<u64>,
-    pub google_id: Option<u128>,
-    pub microsoft_id: Option<u64>,
-    pub apple_id: Option<u64>,
-    pub gitlab_id: Option<u64>,
+    // pub discord_id: Option<u64>,
+    // pub google_id: Option<u128>,
+    // pub microsoft_id: Option<u64>,
```
thesuzerain, line 52: See above

src/models/users.rs
```diff
@@ -147,11 +145,11 @@ impl From<DBUser> for User {
...
             payout_data: None,
             github_id: None,
-            discord_id: None,
-            google_id: None,
-            microsoft_id: None,
-            apple_id: None,
-            gitlab_id: None,
+            // discord_id: None,
+            // google_id: None,
+            // microsoft_id: None,
```
thesuzerain, line 150: See above

thesuzerain (approved):

src/auth/session.rs
Geometrically: it won't be in the API, so no need

src/auth/validate.rs
Geometrically: we won't support cookie auth due to csrf, the frontend supports this auth method

src/auth/validate.rs
Geometrically: i think I'll keep them in the future, for stuff like oauth linking

PR #649, opened 2023-07-08
Sessions Route + Password Auth

Geometrically:
Fixes MOD-492
FIxes MOD-489
Fixes MOD-495


src/auth/flows.rs
```diff
@@ -887,3 +896,155 @@ pub async fn auth_callback(
...
+    new_account
+        .0
+        .validate()
+        .map_err(|err| ApiError::InvalidInput(validation_errors_to_string(err, None)))?;
+
+    if check_turnstile_captcha(&req, &new_account.challenge).await? {
+        return Err(ApiError::Turnstile);
+    }
+
+    if crate::database::models::User::get(&new_account.username, &**pool, &redis)
```
thesuzerain, line 927: We should also do an email collision check here- no duplicate emails (at least against verified emails?))

src/auth/flows.rs
```diff
@@ -887,3 +896,155 @@ pub async fn auth_callback(
...
+        &new_account.password,
+        &[&new_account.username, &new_account.email],
+    )?;
+
+    if score.score() < 3 {
+        return Err(ApiError::InvalidInput(
+            if let Some(feedback) = score.feedback().clone().and_then(|x| x.warning()) {
+                format!("Password too weak: {}", feedback)
+            } else {
+                "Specified password is too weak! Please improve its strength.".to_string()
```
thesuzerain, line 949: It would be good to add a little more information here. Seems like zxcvbn doesnt give info beyond the regressed score but maybe we could add ways as examples so people dont get frustrated ("try using a longer password, or less common words" etc)

src/auth/flows.rs
```diff
@@ -887,3 +896,155 @@ pub async fn auth_callback(
...
+}
+
+#[post("login")]
+pub async fn login_password(
+    req: HttpRequest,
+    pool: Data<PgPool>,
+    redis: Data<deadpool_redis::Pool>,
+    login: web::Json<Login>,
+) -> Result<HttpResponse, ApiError> {
+    if check_turnstile_captcha(&req, &login.challenge).await? {
```
thesuzerain, line 1007: Do we want this to be present on every login attempt? Turnstile is not as obnoxious as some other captchas but it is still indeed another step that some find annoying if not on the 3rd login attempt or smthn

src/auth/flows.rs
Geometrically: They have a suggestions feature, but this is mainly for backend checking. On the frontend we'll have a client side view for this with suggestions.

PR #651, opened 2023-07-10
Scoped PATs

Geometrically:
Fixes MOD-423


src/models/pats.rs
```diff
@@ -0,0 +1,137 @@
...
+    // these scopes cannot be specified in a personal access token
+    pub fn restricted(&self) -> bool {
+        self.contains(
+            Scopes::PAT_CREATE
+                | Scopes::PAT_READ
+                | Scopes::PAT_WRITE
+                | Scopes::PAT_DELETE
+                | Scopes::SESSION_READ
+                | Scopes::SESSION_DELETE
+                | Scopes::USER_AUTH_WRITE,
```
thesuzerain, line 100: probably should also have USER_DELETE on here (as it inherently also does PAT_DELETE anyway)

PR #654, opened 2023-07-13
Verify Email + Reset Password flows

Geometrically:
Fixes MOD-493, MOD-497, MOD-499


src/auth/email/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+    Env(#[from] dotenvy::Error),
+    #[error("Mail Error: {0}")]
+    Mail(#[from] lettre::error::Error),
+    #[error("Address Parse Error: {0}")]
+    Address(#[from] lettre::address::AddressError),
+    #[error("SMTP Error: {0}")]
+    Smtp(#[from] lettre::transport::smtp::Error),
+}
+
+pub fn send_email_raw(to: String, subject: String, body: String) -> Result<(), MailError> {
```
thesuzerain, line 19: Lots of unwraps here in this function. I don't think that any of are guaranteed to be Ok()

src/auth/flows.rs
```diff
@@ -1438,9 +1561,385 @@ pub async fn remove_2fa(
...
+    if let Some(email) = user.email {
+        let changed = if update_password.is_some() {
+            "changed"
+        } else {
+            "removed"
+        };
+
+        send_email(
+            email,
+            &format!("Password {}", changed),
```
thesuzerain, line 1753: More of a general thing, I don't think we need to send emails for things like password changes (removals, maybe). Is it just for extra safety? I don't really remember any other sites doing this

Minenash: I'm trying to think of a site where I don't get an email after changing my password (assuming they have my email in the first place)

PR #659, opened 2023-07-18
Finish authentication

Geometrically:


thesuzerain (approved):

src/auth/flows.rs
```diff
@@ -1930,6 +2007,34 @@ pub async fn verify_email(
     }
 }
 
+#[post("email/subscribe")]
```
thesuzerain, line 2010: we should have an email/unsubscribe as well (or a delete("email/subscribe"), or some way to unsubscribe)

thesuzerain (approved):

src/auth/flows.rs
Geometrically: done through beehiiv

PR #671, opened 2023-08-07
remove Fabric 'source file' validation

Jamalam360:
discussed on discord

### DIFF
--- a/src/validate/fabric.rs
+++ b/src/validate/fabric.rs
@@ -36,15 +36,6 @@ impl super::Validator for FabricValidator {
             ));
         }
 
-        if !archive
-            .file_names()
-            .any(|name| name.ends_with("refmap.json") || name.ends_with(".class"))
-        {
-            return Ok(ValidationResult::Warning(
-                "Fabric mod file is a source file!",
-            ));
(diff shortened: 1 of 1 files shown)


Jamalam360:
side note: `ValidationResult::Warning` seems like something that shouldn't prevent publishing. On the frontend it's displayed as a red error box

ThatGravyBoat:
Is there a reason why this is only being removed for fabric and not all loaders?

Geometrically:
we don't do this on other loaders

ThatGravyBoat:
@Geometrically 
> we don't do this on other loaders

You do it in every other mod loader:
https://github.com/modrinth/labrinth/blob/master/src/validate/forge.rs#L39
https://github.com/modrinth/labrinth/blob/master/src/validate/forge.rs#L90
https://github.com/modrinth/labrinth/blob/master/src/validate/quilt.rs#L38 (This one is actually a 1:1 of the fabric check)

lukebemish:
The forge validator is particularly insidious; a forge mod need not have a mods.toml file to be valid, even on new versions!

PR #698, opened 2023-08-30
Allow BungeeCord plugins that use bungee.yml

### LINKED ISSUE #687: Upload Bungeecord plugin but failed with error
### Describe the bug

Upload Bungeecord plugin but failed with error 'No plugin.yml found in the file'

But in fact, Bungeecord plugin didn't has 'plugin.yml' in the jar file. It only has 'bungee.yml'.

And for Velocity plugins, it only has 'velocity-plugin.json'

### Steps to reproduce

Add new project

Upload a Bungeecord or Velocity plugin.

Get the error.

### Expected behavior

Upload successfully.

### Additional context

I don't know if it's a UI problem or an API problem
So I posted the same issue at https://github.com/modrinth/knossos/issues/1299

Konicai:
First time Rust user here. I'm under the impression that all of the `VALIDATORS` in rust.rs are tested to see if they match the given primary file upload? It seems like PluginYmlValidator was matching for bungeecord/waterfall but then subsequently denying the upload because it doesn't cover the bungeecord usage of `bungee.yml` as described in #687.

bungeecord/waterfall was removed as a supported platform of PluginYmlValidator, as there is already BungeeCordValidator. I believe this should fix the issue.
Also, support for `plugin.yml` was added to BungeeCordValidator (although that wasn't the current source of this issue).

As discussed on Discord, I haven't tested this PR.

Closes #687

### DIFF
--- a/src/validate/plugin.rs
+++ b/src/validate/plugin.rs
@@ -18,7 +18,7 @@ impl super::Validator for PluginYmlValidator {
             "bukkit",
             "spigot",
             "paper",
-            "purpur"
+            "purpur",
         ]
     }
 
--- a/src/validate/plugin.rs
+++ b/src/validate/plugin.rs
@@ -18,9 +18,7 @@ impl super::Validator for PluginYmlValidator {
             "bukkit",
             "spigot",
             "paper",
-            "purpur",
-            "bungeecord",
-            "waterfall",
+            "purpur"
         ]
     }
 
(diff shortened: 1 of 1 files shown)


Konicai:
seems like the lint warnings are unrelated to this PR?

triphora:
Yeah, no need to worry about that

src/validate/plugin.rs
```diff
@@ -18,9 +18,7 @@ impl super::Validator for PluginYmlValidator {
             "bukkit",
             "spigot",
             "paper",
-            "purpur",
-            "bungeecord",
-            "waterfall",
+            "purpur"
```
triphora, line 21:
```suggestion
            "purpur",
```

triphora (approved):

PR #707, opened 2023-09-10
Add NeoForge modpack and autoupdater support

triphora:


Geometrically (approved):

LostLuma:
Hello, I'm not sure where the best place to ask for it is but this feature is currently missing from the documentation, making it seem like it doesn't exist without further searching.

PR #712, opened 2023-09-19
Organizations

thesuzerain:


src/models/organizations.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// The associated team of the organization
+    pub team_id: TeamId,
+    /// The name of the organization
+    pub name: String,
+    /// The description of the organization
+    pub description: String,
+    /// The website url of the organization
+    pub website_url: Option<String>,
+    /// The discord url of the organization
+    pub discord_url: Option<String>,
```
Geometrically, line 30: I think we should restructure this to an HashMap of an enum and string. this is the plan for other link structures, but since this is new we should do it here

src/routes/v2/organizations.rs
```diff
@@ -0,0 +1,1153 @@
...
+        .into_iter()
+        .next();
+    let members_data = if let Some(member_data) = member_data {
+        vec![crate::models::teams::TeamMember::from(
+            member_data,
+            user_data.clone(),
+            false,
+        )]
+    } else {
+        vec![]
```
Geometrically, line 217: this should error not soft fail

Geometrically (changes requested):

src/database/models/team_item.rs
```diff
@@ -520,4 +649,31 @@ impl TeamMember {
...
+
+    // Gets all required members for checking permissions of an action on a project
+    // - project team member (a user's membership to a given project)
+    // - organization team member (a user's membership to a given organization that owns a given project)
+    // - organization (the organization that owns a given project)
+    pub async fn get_for_project_permissions<'a, 'b, E>(
+        project: &Project,
+        user_id: UserId,
+        executor: E,
+    ) -> Result<(Option<Self>, Option<Self>, Option<Organization>), super::DatabaseError>
```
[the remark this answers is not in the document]
thesuzerain: The issue is for the most part we do need all three of these: 

- the team member of this team (to check if we are a member of the team)
- the organization team if there is one (to check if we are a member of the team's 
- Organization, and if we are the owner of said organization) 
- and the organization itself (to check the default permissions of the organization)
We could reognize this into one query that returns all three- but these three are currently cached with redis which is pretty nice

src/database/models/organization_item.rs
```diff
@@ -0,0 +1,380 @@
...
+
+    /// The name of the organization
+    pub name: String,
+
+    /// The description of the organization
+    pub description: String,
+
+    /// Default project permissions for associated projects
+    pub default_project_permissions: ProjectPermissions,
+
```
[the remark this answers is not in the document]
thesuzerain: see disc

src/database/models/organization_item.rs
```diff
@@ -0,0 +1,380 @@
...
+            }
+        }
+
+        Ok(found_organizations)
+    }
+
+    // Gets organization associated with a project ID, if it exists and there is one
+    pub async fn get_associated_organization_project_id<'a, 'b, E>(
+        project_id: ProjectId,
+        exec: E,
```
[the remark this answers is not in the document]
thesuzerain: An org is associated with a project when an org owns that project\
This gets the org that is owned by project ID if there is one

src/database/models/team_item.rs
```diff
@@ -474,7 +552,53 @@ impl TeamMember {
...
+            Ok(None)
+        }
+    }
+
+    pub async fn get_from_user_id_organization<'a, 'b, E>(
+        id: OrganizationId,
+        user_id: UserId,
+        executor: E,
+    ) -> Result<Option<Self>, super::DatabaseError>
+    where
```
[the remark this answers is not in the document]
thesuzerain: Its equivalent to get_from_user_id_project but for organization teams

src/database/models/organization_item.rs
```diff
@@ -0,0 +1,361 @@
...
+
+#[derive(Deserialize, Serialize, Clone, Debug)]
+/// An organization of users who together control one or more projects and organizations.
+pub struct Organization {
+    /// The id of the organization
+    pub id: OrganizationId,
+
+    /// The slug of the organization
+    pub slug: String,
+
```
Geometrically, line 25: org slugs and names should be the same, like usernames. it shouldn't be an additional field

src/database/models/organization_item.rs
```diff
@@ -0,0 +1,361 @@
...
+    pub name: String,
+
+    /// The description of the organization
+    pub description: String,
+
+    /// Default project permissions for associated projects
+    pub default_project_permissions: ProjectPermissions,
+
+    /// Any associated urls for the organization
+    pub urls: HashMap<String, String>,
```
Geometrically, line 39: i think this should be an enum with possible allowed links

Geometrically: might want to talk on vc on how we structure this

src/database/models/organization_item.rs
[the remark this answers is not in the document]
Geometrically: when would this be used though?

src/database/models/organization_item.rs
```diff
@@ -0,0 +1,361 @@
...
+        id: OrganizationId,
+        transaction: &mut sqlx::Transaction<'_, sqlx::Postgres>,
+        redis: &deadpool_redis::Pool,
+    ) -> Result<Option<()>, super::DatabaseError> {
+        let project = Self::get_id(id, &mut *transaction, redis).await?;
+
+        if let Some(organization) = project {
+            Organization::clear_cache(id, Some(organization.slug), redis).await?;
+
+            sqlx::query!(
```
Geometrically, line 304: make sure to remove associated projects here. also if you ever link data to something else in the database it should be removed in the delete function (just for future ref)

src/database/models/team_item.rs
```diff
@@ -27,44 +28,48 @@ impl TeamBuilder {
...
             ",
             team.id as TeamId,
+            match team.association {
+                TeamAssociationId::Project(pid) => Some(pid.0),
+                TeamAssociationId::Organization(_) => None,
+            },
+            match team.association {
+                TeamAssociationId::Project(_) => None,
+                TeamAssociationId::Organization(oid) => Some(oid.0),
+            },
```
Geometrically, line 49: does team association need to be kept track of in the api? don't see the point. might be wrong tho

src/database/models/team_item.rs
```diff
@@ -27,44 +28,48 @@ impl TeamBuilder {
...
-                team_member.role,
-                team_member.permissions.bits() as i64,
-                team_member.accepted,
-                team_member.payouts_split,
-                team_member.ordering,
+                team_member_id as TeamMemberId,
+                team.id as TeamId,
+                member.user_id as UserId,
+                member.role,
+                match member.permissions {
```
Geometrically, line 65: perms should be split into 2 fields as discussed

src/database/models/team_item.rs
```diff
@@ -78,17 +83,68 @@ impl TeamBuilder {
...
     pub id: TeamId,
+    // The association of the team: either a project or an organization
+    pub association: TeamAssociationId,
+}
+
+#[derive(Deserialize, Serialize, Clone, Debug, Copy)]
+pub enum TeamAssociationId {
+    Project(ProjectId),
+    Organization(OrganizationId),
+}
```
Geometrically, line 94: think this should be removed from the team/is not needed

src/database/models/team_item.rs
```diff
@@ -78,17 +83,68 @@ impl TeamBuilder {
...
+    }
 }
 
 /// A member of a team
-#[derive(Deserialize, Serialize)]
+#[derive(Deserialize, Serialize, Clone, Debug)]
 pub struct TeamMember {
     pub id: TeamMemberId,
     pub team_id: TeamId,
+    pub association_id: TeamAssociationId,
```
Geometrically, line 141: when is this field used?

src/database/models/team_item.rs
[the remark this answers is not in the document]
Geometrically: yeah but when is it used

src/database/models/team_item.rs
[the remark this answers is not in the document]
Geometrically: when is this called? is it on every get function or only on modifying functions (like deleting a project)

src/models/teams.rs
```diff
@@ -21,10 +21,17 @@ pub struct Team {
     pub members: Vec<TeamMember>,
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug)]
+#[serde(untagged)]
+pub enum Permissions {
+    Project(ProjectPermissions),
+    Organization(OrganizationPermissions),
+}
+
```
Geometrically, line 30: these should be separated into 2 different fields

src/routes/v2/organizations.rs
```diff
@@ -0,0 +1,1038 @@
...
+            ",
+            slug_organization_id.0 as i64
+        )
+        .fetch_one(&mut *transaction)
+        .await
+        .map_err(|e| CreateError::DatabaseError(e.into()))?;
+
+        if results.exists.unwrap_or(false) {
+            return Err(CreateError::SlugCollision);
+        }
```
Geometrically, line 105: you should just be able to use the organization get func here right>?

src/routes/v2/organizations.rs
```diff
@@ -0,0 +1,1038 @@
...
+            new_organization.slug
+        )
+        .fetch_one(&mut *transaction)
+        .await
+        .map_err(|e| CreateError::DatabaseError(e.into()))?;
+
+        if results.exists.unwrap_or(false) {
+            return Err(CreateError::SlugCollision);
+        }
+    }
```
Geometrically, line 122: see above

src/routes/v2/organizations.rs
```diff
@@ -0,0 +1,1038 @@
...
+
+    // Gets the user database data
+    let user_data =
+        crate::database::models::User::get_id(current_user.id.into(), &mut transaction, &redis)
```
Geometrically, line 126: why is this query needed? you already have the user

src/routes/v2/teams.rs
```diff
@@ -65,6 +64,69 @@ pub async fn team_members_get_project(
...
+            TeamMember::get_from_team_full(organization.team_id, &**pool, &redis).await?;
+        let users = crate::database::models::User::get_many_ids(
+            &members_data.iter().map(|x| x.user_id).collect::<Vec<_>>(),
+            &**pool,
+            &redis,
+        )
+        .await?;
+
+        let user_id = current_user.as_ref().map(|x| x.id.into());
+
```
Geometrically, line 129: i dont think this route is needed given you already include it in the org reponse

src/routes/v2/teams.rs
```diff
@@ -65,6 +64,69 @@ pub async fn team_members_get_project(
 
         let user_id = current_user.as_ref().map(|x| x.id.into());
```
Geometrically, line 65: does this now return the orgs/team members merged team? if not it should

Geometrically (changes requested):

src/routes/v2/teams.rs
thesuzerain: Yeah but removing it would remove the ability to check members for project teams as well, as this is a team route. 
I can remove its organization-specific functionality and keep the project one, but I think that might confuse users more ("if orgs use teams, why dont team routes work for them)

src/routes/v2/organizations.rs
thesuzerain: I was using it for the db Team Member -> models Team Member function which requires db Users and not models Users. I can create a version of that function that uses the models version though as it looks like that doesnt use the db version for anything other than converting it

migrations/20230913024611_organizations.sql
```diff
@@ -0,0 +1,27 @@
...
+    id bigint PRIMARY KEY,
+    name varchar(255) NOT NULL,
+    slug varchar(255) NOT NULL,
+    description text NOT NULL,
+    created_at timestamp NOT NULL DEFAULT now(),
+    updated_at timestamp NOT NULL DEFAULT now(),
+    default_project_permissions bigint NOT NULL DEFAULT 0,
+    team_id bigint NOT NULL REFERENCES teams(id) ON UPDATE CASCADE,
+
+    urls varchar(2048) NOT NULL,
```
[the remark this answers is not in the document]
thesuzerain: This is a serialized hashmap for any urls they want to include 
(discord, wiki, etc)

migrations/20230913024611_organizations.sql
```diff
@@ -0,0 +1,27 @@
...
+
+);
+
+ALTER TABLE mods ADD COLUMN organization_id bigint NULL REFERENCES organizations(id) ON DELETE SET NULL;
+
+-- Allows getting of organiztion/project from team
+ALTER TABLE teams ADD COLUMN organization_id bigint NULL REFERENCES organizations(id) ON DELETE SET NULL;
+ALTER TABLE teams ADD COLUMN project_id bigint NULL REFERENCES mods(id) ON DELETE SET NULL;
+
+ALTER TABLE team_members ALTER COLUMN permissions DROP NOT NULL;
```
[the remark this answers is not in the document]
thesuzerain: This was a holdover from the schema where only one of permissions or organization_permissions would be true

src/database/models/organization_item.rs
thesuzerain: I set it up like donation links- we can discuss monday if you want it a specific different way

migrations/20230913024611_organizations.sql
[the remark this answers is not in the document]
thesuzerain: See urls below

src/database/models/organization_item.rs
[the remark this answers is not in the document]
thesuzerain: its used directly in get_for_project_permissions, but also in a lot of places where get_for_project_permissions is not necessary (to do all checks within) but we still need to get what-organization-information from a project (getting org owner, getting org team, etc)

src/database/models/team_item.rs
thesuzerain: Kept the structure as its useful, but no longer in the db as discussed on disc

src/database/models/team_item.rs
thesuzerain: Kept the structure as its useful, but no longer in the db as discussed on disc

src/database/models/team_item.rs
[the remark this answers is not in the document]
thesuzerain: Should only be on modifying functions, not on gets. But that being said, the gets within  should be cached anyway

src/routes/v2/organizations.rs
thesuzerain: Yeah, Im realizing I was copying behaviour in project_create that wasnt necessary (as it was created before redis was added). Maybe that should be changed too- what do you think? I can revisit project_create my testing PR

Geometrically (changes requested):

Geometrically (approved):

PR #719, opened 2023-09-24
Tests

thesuzerain:
Still in draft for github actions testing alongside actual test writing.
Theoretically in the future, adds an automatic cargo test library to labrinth


src/database/models/user_item.rs
```diff
@@ -371,32 +352,26 @@ impl User {
...
         user_ids: &[(UserId, Option<String>)],
-        redis: &deadpool_redis::Pool,
+        redis: &RedisPool,
     ) -> Result<(), DatabaseError> {
-        let mut redis = redis.get().await?;
-        let mut cmd = cmd("DEL");
-
         for (id, username) in user_ids {
-            cmd.arg(format!("{}:{}", USERS_NAMESPACE, id.0));
+            redis.delete(USERS_NAMESPACE, id.0).await?;
```
OmegaJak, line 358: This redis delete was previously a single operation, now it's `user_ids.len()` operations. I think we want the batching wherever we can get it.

Maybe we could hava `delete_many` method on the `RedisPool` that accepts a slice of tuples of something like `(String, impl Display)`? Though in this case we actually have two types - UserIds and Option<String>... so maybe it should actually take an `impl IntoIterator<Item = (String, String)>`?

src/database/redis.rs
```diff
@@ -0,0 +1,110 @@
...
+        ids: Vec<T1>,
+    ) -> Result<Vec<Option<R>>, DatabaseError>
+    where
+        T1: Display,
+        R: FromRedisValue,
+    {
+        let mut redis_connection = self.pool.get().await?;
+        let res = cmd("MGET")
+            .arg(
+                ids.iter()
```
OmegaJak, line 88: Since the only way we're using `ids` is to `.iter()`, we could allow its type to be `ids: impl IntoIterator<Item = T1>`. This would allow us to avoid unnecessarily allocating a `Vec` in the places where we're calling `multi_get`

src/database/models/pat_item.rs
```diff
@@ -237,42 +218,37 @@ impl PersonalAccessToken {
...
             return Ok(());
         }
 
-        let mut redis = redis.get().await?;
-        let mut cmd = cmd("DEL");
-
         for (id, token, user_id) in clear_pats {
             if let Some(id) = id {
-                cmd.arg(format!("{}:{}", PATS_NAMESPACE, id.0));
+                redis.delete(PATS_NAMESPACE, id.0).await?;
```
OmegaJak, line 242: Same concern about unbatching here

src/models/pats.rs
```diff
@@ -104,15 +104,15 @@ bitflags::bitflags! {
         const ORGANIZATION_DELETE = 1 << 38;
 
         const ALL = 0b111111111111111111111111111111111111111;
-        const NOT_RESTRICTED = 0b1111111100000011111111111111100111;
+        const NOT_RESTRICTED = 0b1111111110000000111111111111111111100111;
```
OmegaJak, line 107: Well this is kinda terrifying... glad it's tested now

Do we care that `NOT_RESTRICTED` has an extra 1? It specifies 40 bits where ALL specifies 39.

src/routes/v2/collections.rs
```diff
@@ -231,7 +231,7 @@ pub async fn collection_edit(
     let result = database::models::Collection::get(id, &**pool, &redis).await?;
 
     if let Some(collection_item) = result {
-        if !is_authorized_collection(&collection_item, &user_option).await? {
+        if collection_item.user_id != user.id.into() && !user.role.is_mod() {
```
OmegaJak, line 234: Why aren't we using the common method any more? I don't think we want to be introducing more duplication if we can help it

src/routes/v2/version_file.rs
```diff
@@ -21,7 +22,8 @@ pub fn config(cfg: &mut web::ServiceConfig) {
             .service(delete_file)
             .service(get_version_from_hash)
             .service(download_version)
-            .service(get_update_from_hash),
+            .service(get_update_from_hash)
+            .service(get_projects_from_hashes), // TODO: confirm this should be added
```
OmegaJak, line 26: This something to discuss in this PR?

Cargo.toml
```diff
@@ -18,6 +18,7 @@ actix-multipart = "0.6.0"
 actix-cors = "0.6.4"
 actix-ws = "0.2.5"
 actix-files = "0.6.2"
+actix-http = "3.4.0"
```
OmegaJak, line 21: It looks like this is only used in the tests, so we should put it into the [\[dev-dependencies\]](https://doc.rust-lang.org/rust-by-example/testing/dev_dependencies.html) section to keep it out of the main binary.

tests/pats.rs
```diff
@@ -0,0 +1,292 @@
...
+            .to_request();
+        let resp = test_env.call(req).await;
+        assert_eq!(
+            resp.status().as_u16(),
+            if scope.restricted() { 400 } else { 204 }
+        );
+    }
+
+    // Cleanup test db
+    test_env.cleanup().await;
```
OmegaJak, line 291: It would be great if we could find a way to do this on `Drop` of the test_env, as we're almost guaranteed to forget to call cleanup at some point when writing future tests. The fact that this is async substantially complicates things though. Here are some possible ways to do it that I found: https://stackoverflow.com/a/74511028, https://stackoverflow.com/questions/71541765/rust-async-drop

tests/pats.rs
```diff
@@ -0,0 +1,292 @@
...
+
+// Full pat test:
+// - create a PAT and ensure it can be used for the scope
+// - ensure access token is not returned for any PAT in GET
+// - ensure PAT can be patched to change scopes
+// - ensure PAT can be patched to change expiry
+// - ensure expired PATs cannot be used
+// - ensure PATs can be deleted
+#[actix_rt::test]
+pub async fn pat_full_test() {
```
OmegaJak, line 20: Just a general comment, not looking for an actual change - I hope that we can progress, with time, towards smaller tests that get closer to testing a small slice of functionality. Smaller tests makes it easier to diagnose test failures at a glance. If one of the early asserts in this test fails, then we get no information about the state of the other things it's testing.

However, given the current overhead of setting up for a test, and the sequential nature of these operations, this seems like a great place to start.

OmegaJak (commented):
Love it! Can't wait to use it

src/database/redis.rs
thesuzerain: Solid, yeah good point

src/models/pats.rs
thesuzerain: Ill do a recount. I think they should be the same, but the test should be able to catch it so I'll look at that again

src/routes/v2/collections.rs
thesuzerain: The issue here is that is_authorized_collection doesnt actually check if its a collection that *belongs* to you, only whether its visible to you. This was one of the bugs I found.

tests/pats.rs
thesuzerain: This is on my list for sure, I gave it a go but it was frustrating. We have a similar async drop trait in theseus (for loading bars) but the difference is for that one the runtime continues, whereas the async runtime drops at the end of the test here (and not necessarily cleaning up the database). I tried a couple solutions but they didnt seem to work, so its something on my list to come back to.

Ideally, we drop this manual database creation-dropping system entirely and use #[sqlx::test], but I had similar adjacent problems when I use sqlx::test instead of actix::test.

tests/pats.rs
thesuzerain: Yeah reasonable.  Theres something of a trade off with the test size that is worth playing with in the future

src/models/pats.rs
OmegaJak: It's apparent if you line them up in something monospaced like VSCode: 
image

src/database/redis.rs
```diff
@@ -0,0 +1,110 @@
...
+        namespace: &str,
+        id: T1,
+        data: T2,
+        expiry: Option<i64>,
+    ) -> Result<(), DatabaseError>
+    where
+        T1: Display,
+        T2: ToRedisArgs,
+    {
+        let mut redis_connection = self.pool.get().await?;
```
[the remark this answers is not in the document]
thesuzerain: This was something I thought a little about, but I ended up not really worrying about it. Most areas that call redis currently only do 1-2 commands on it anyway per connection, so I don't really think it would affect that much either positively or negatively. 

I don't think it would be entirely too difficult to set it up that way, aside from some mild structure complication (ie: another struct for RedisConnection) if we think it would be an issue. @Geometrically thoughts?

tests/pats.rs
OmegaJak: Dang yeah, figured. Something we can try to get in the future.

src/routes/v2/collections.rs
OmegaJak: Could we make a new method that does it then? Don't love that this line is copy/pasted 4 times across this file...

src/database/redis.rs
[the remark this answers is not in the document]
OmegaJak: Yeah, I'm inclined to agree that it's fine, especially since when we have two we're usually doing a slow operation (database call) in between.

tests/scopes.rs
Geometrically, line 1: The structure here seems a bit weird to me. I think it would be better to just test the scopes as part of other tests and have those in respective files. Since this is testing the routes anyways basically I think that structure would be better. Totally open to both

tests/files/dummy_data.sql
Geometrically, line 1: Also a little concerned with this- the database structure will change pretty frequently so I can see this being pretty annoying to update. Besides users, couldn't the other items be created through the API? That way nothing ever ends up in ilegal state while testing others

tests/common/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+        file_host.clone(),
+        maxmind_reader.clone(),
+    )
+}
+
+// This is so that env vars not used immediately don't panic at runtime
+// Currently, these are the same as main.rs ones.
+// TODO: go through after all tests are created and remove any that are not used
+// Low priority as .env file should include all of these anyway
+fn check_test_vars() -> bool {
```
Geometrically, line 47: doesn't labrinth's init function already call this (which is called from the tests). if not, couldn't you just reuse the labrinth function? also might be annoying to add env variables here as well when you add a new one

tests/common/environment.rs
```diff
@@ -0,0 +1,208 @@
...
+        &self,
+        req: actix_http::Request,
+    ) -> std::pin::Pin<
+        Box<dyn std::future::Future<Output = Result<ServiceResponse, actix_web::Error>>>,
+    > {
+        Box::pin(self.call(req))
+    }
+}
+
+// A reusable test type that works for any scope test testing an endpoint that:
```
Geometrically, line 72: I think the 2 items below this (scopetest and create pat) probably belong in a different file or `scopes.rs` and `pats.rs` respectively

Geometrically (changes requested):

tests/scopes.rs
thesuzerain: Thats fair- if its all the same I think I'll keep it this way for now given that we dont have files for all the routes yet. As we make rs files ill port them over

src/models/pats.rs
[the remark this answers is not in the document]
thesuzerain: Sure. I can get rid of this test then too because it becomes trivial (does this variable equal itself basically)

tests/common/mod.rs
thesuzerain: Yeah, I thought there may be a case where they may be different (ie: ability to run tests without env vars). But I guess it doesn't matter so much- .env exists

src/routes/v2/collections.rs
thesuzerain: Fair enough!

src/database/models/user_item.rs
OmegaJak: Looks like the clear_cache methods that use `delete_many` now still aren't as batched as they were - many are still calling `redis.delete_many` inside of their for loops. I think instead they should be building an iterator of the keys to delete, then calling `delete_many` once at the end.

Also it doesn't appear to be used in `version_item.rs`, `user_item.rs` are batched at all

OmegaJak: Looks good!

OmegaJak (approved):

Geometrically (approved):

PR #721, opened 2023-09-30
Make use of new `Fabric-Jar-Type` entry from Loom

triphora:
I'm not entirely sure if this is necessary to add, but this is here in case we do want to begin making use of it.


BasiqueEvangelist:
modrinth/.github#176 :^)
Would be nice to also automatically expose sources jars the proper way on maven

triphora:
Closing since I am not gonna be the one to implement this, if someone else wants to pick it up feel free.

PR #723, opened 2023-10-03
Add redis caching to getting user notifications and projects [MOD-540]

OmegaJak:
Specifically:
- /user/{id|username}/notifications
- /user/{id|username}/projects

both of these now cache their results in redis.

Cases I could think of that the user projects cache needs invalidation:
- Team member joined (clear team member's cache)
- Team member removed (clear team member's cache)
- Project created (clear cache of creating user)
- Project removed (clear cache of all team members)

Cases I could think of that the user notifications cache needs invalidation:
- Notification added
- Notification removed
- Notification read


OmegaJak:
Draft because I haven't tested this much yet and still need to. I'm hoping to write automated tests around the invalidation conditions using #719

src/database/models/notification_item.rs
```diff
@@ -30,52 +36,64 @@ impl NotificationBuilder {
...
+        Notification::insert_many(&notifications, transaction, redis).await?;
+
         Ok(())
     }
 }
 
 impl Notification {
-    pub async fn insert(
-        &self,
+    pub async fn insert_many(
```
OmegaJak, line 70: I did this for a couple reasons:
- The only caller of `insert` was `NotificationBuilder::insert_many`
- We should be batching our Postgres inserts using this: https://github.com/launchbadge/sqlx/blob/main/FAQ.md#how-can-i-bind-an-array-to-a-values-clause-how-can-i-do-bulk-inserts
- This allows me to batch the delete from redis while maintaining only a single code path. If the single insert had been allowed to continue, I could have done the cache clearing up at `NotificationBuilder::insert_many`, but then that would have risked `Notification::insert` being called and not clearing the cache

src/database/models/user_item.rs
```diff
@@ -298,13 +298,27 @@ impl User {
...
+        let mut redis = redis.get().await?;
+
+        let cached_projects = cmd("GET")
+            .arg(format!("{}:{}", USERS_PROJECTS_NAMESPACE, user_id.0))
+            .query_async::<_, Option<String>>(&mut redis)
+            .await?
+            .and_then(|x| serde_json::from_str::<Vec<ProjectId>>(&x).ok());
+
+        if let Some(projects) = cached_projects {
+            log::info!("Returning cached!");
```
thesuzerain, line 317: I wouldn't mind logging these kinds of things but feels weird to only have the log for this one

src/database/models/notification_item.rs
thesuzerain: Yup this makes sense to me. I would also agree reformatting bulk inserts should be its own PR

src/database/models/user_item.rs
OmegaJak: Whoops, good catch! That was there just for my own reassurance during testing. Didn't intend for it to stay. Will remove

src/database/models/notification_item.rs
OmegaJak: Per discussion with Jai, I went ahead and did it. Completely untested, intending to test with #719

tests/common/dummy_data.rs
```diff
@@ -160,31 +167,31 @@ pub async fn add_project_alpha(test_env: &TestEnvironment) -> (Project, Version)
     (project, version)
 }
 
-pub async fn add_project_beta(test_env: &TestEnvironment) -> (Project, Version) {
+pub async fn add_public_dummy_project(
```
thesuzerain, line 170: I think the one part I disagree with is having `add_public_dummy_project` as a function to be used in tests. Im all for a helper function for easy creation of projects (like you are doing for `TestEnvironment`) , but for dummy data specifically im not sure.

 Mostly because the speedup changes involve putting the dummy data into a template db, meaning that you generate the dummy data in the first test and never again, and it doesn't vary test to test (so it feels wrong to generate dummy data in tests).

We could instead do a `add_public_project` on `TestEnvironment` or something similar- what do you think

tests/common/environment.rs
```diff
@@ -54,6 +61,111 @@ impl TestEnvironment {
...
+            .add_user_to_team(
+                &self.dummy.as_ref().unwrap().alpha_team_id,
+                FRIEND_USER_ID,
+                USER_USER_PAT,
+            )
+            .await;
+        assert_status(resp, StatusCode::NO_CONTENT);
+    }
+
+    pub async fn remove_project(&self, project_slug_or_id: &str, pat: &str) -> ServiceResponse {
```
thesuzerain, line 76: I think the helper functions here on `TestEnvironment`  is a really good idea for avoiding duplicated code in setting up tests.

This is probably going to change a lot though soon once we have v3 routes though. Maybe we can do two `TestEnvironments` (or a struct in the fields of `TestEnvironment` that contains the helpers) that implement a common trait?

tests/common/dummy_data.rs
thesuzerain: To be clear I think having an easy create-project function is good but associating with the dummy data is what I disagree with

tests/common/dummy_data.rs
```diff
@@ -75,26 +86,34 @@ pub async fn add_dummy_data(test_env: &TestEnvironment) -> DummyData {
...
+            "title": "Test Project Beta",
+            "slug": "beta",
             "description": "A dummy project for testing with.",
-            "body": "This project is approved, and versions are listed.",
+            "body": "This project is not-yet-approved, and versions are draft.",
             "client_side": "required",
             "server_side": "optional",
             "initial_versions": [{
-                "file_parts": ["dummy-project-alpha.jar"],
+                "file_parts": ["dummy-project-beta.jar"],
```
thesuzerain, line 104: These may as well use the new DummyJarBeta constants as well

thesuzerain (changes requested):

tests/common/dummy_data.rs
OmegaJak: Oh good point - the name felt weird anyway. I like the idea of moving to to `TestEnvironment`, will do.

tests/common/dummy_data.rs
OmegaJak: Good idea, will do

tests/common/environment.rs
OmegaJak: Thanks!

And after some separate discussion I'll be changing it to put the helpers into a separate struct that's accessible through the TestEnvironment.

tests/common/dummy_data.rs
OmegaJak: Didn't go as far as making the beta create project use the function yet because of the differences between them, but we should try and get that at some point. It uses the DummyJar enum now, though.

thesuzerain (approved):

tests/common/dummy_data.rs
thesuzerain: Sounds good, I can have a look at this in my PR as I'm updating the dummy data code anyway

Geometrically (approved):

PR #724, opened 2023-10-06
Add /metrics endpoint for Prometheus [MOD-545]

OmegaJak:


Cargo.toml
```diff
@@ -18,6 +18,7 @@ actix-multipart = "0.6.0"
 actix-cors = "0.6.4"
 actix-ws = "0.2.5"
 actix-files = "0.6.2"
+actix-web-prom = "0.7.0"
```
OmegaJak, line 21: All of the changes besides this were just auto-formats, do we mind?

src/main.rs
```diff
@@ -101,6 +107,7 @@ async fn main() -> std::io::Result<()> {
     // Init App
     HttpServer::new(move || {
         App::new()
+            .wrap(prometheus.clone())
```
OmegaJak, line 110: Didn't play nice with the compression middleware, so I had to put it above it. Didn't investigate why, could if we want.

Geometrically (approved):

PR #726, opened 2023-10-09
Batch inserts [MOD-555]

OmegaJak:
Convert all instances of inserts being done inside for loops that I could find (excluding doubly nested or other extra gnarly scenarios) into using batch inserts in the way [currently recommended by Sqlx](https://github.com/launchbadge/sqlx/blob/main/FAQ.md#how-can-i-bind-an-array-to-a-values-clause-how-can-i-do-bulk-inserts).

Most of the changes are pretty straightforward transformations where I replaced the for loop with an iterator statement that produces a vec per column. A few others were a bit more involved:
- Payouts couldn't be fully transformed because it's also doing updates inside its for loop. That part hasn't changed. Now we just push to some vecs as we iterate and do a bulk insert at the end
- Clickhouse was asking for some deduplication, so I refactored a bit to abstract commonalities between the different analytics. Because this was not the standard postgres refactor, and because it's relatively isolated, I wrote some tests around this.
- There was some very substantial duplication in `routes/v2/projects.rs` around category insertion, so I refactored commonalities there.

I also didn't touch notification inserts is pending in #723.

Not yet tested outside of existing route tests and the tests around clickhouse.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1272,6 +1272,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "derive-new"
+version = "0.5.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3418329ca0ad70234b9735dc4ceed10af4df60eff9c8e7b06cb5e520d92c3535"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 1.0.109",
+]
+
+[[package]]
 name = "derive_builder"
 version = "0.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2246,6 +2257,7 @@ dependencies = [
  "color-thief",
  "dashmap",
  "deadpool-redis",
+ "derive-new",
  "dotenvy",
  "env_logger",
  "flate2",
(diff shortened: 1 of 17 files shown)


tests/files/dummy_data.sql
```diff
@@ -27,10 +27,20 @@ INSERT INTO loaders (id, loader) VALUES (1, 'fabric');
...
+    (5, 'magic', 1),
+    (6, 'mobs', 1),
+    (7, 'optimization', 1);
 
-INSERT INTO categories (id, category, project_type) VALUES (4, 'combat', 2);
-INSERT INTO categories (id, category, project_type) VALUES (5, 'decoration', 2);
-INSERT INTO categories (id, category, project_type) VALUES (6, 'economy', 2);
\ No newline at end of file
+INSERT INTO categories (id, category, project_type) VALUES
+    (8, 'combat', 2),
```
thesuzerain, line 40: if we forsee adding more testing categories in the future it might be worth starting this second set at 10 or 20. If you think this is enough, should be OK

thesuzerain (approved):

tests/files/dummy_data.sql
OmegaJak: Good suggestion, I made the next ones start at 101

Geometrically (approved):

PR #729, opened 2023-10-13
More tests

thesuzerain:
Fixes MOD-556


tests/common/permissions.rs
```diff
@@ -0,0 +1,992 @@
...
+        .get_organization_members(organization_id, user_pat)
+        .await;
+    let permissions = if resp.status().as_u16() == 200 {
+        let value: serde_json::Value = test::read_body_json(resp).await;
+        value
+            .as_array()
+            .unwrap()
+            .iter()
+            .find(|member| member["user"]["id"].as_str().unwrap() == user_id)
+            .map(|member| member["organization_permissions"].as_u64().unwrap())
```
Geometrically, line 985: for these ones I think it would be better to put the permissions test of each object in it's own respective file. i think it would be more organized and makes more sense to me

Geometrically (changes requested):
For my comment on the permissions test, I'm thinking we apply that to all of the permissions tests (move them to the respective object file which is being modified!)

src/models/teams.rs
```diff
@@ -59,11 +59,16 @@ impl ProjectPermissions {
         }
 
         if let Some(member) = project_team_member {
+            if member.accepted {
+                return Some(member.permissions);
+            }
             return Some(member.permissions);
```
OmegaJak, line 65: As written, this change has no effect on the behavior of the code - it just always returns `Some(member.permissions)`. Did you maybe intend to only return that when the member has accepted?

OmegaJak:
Looking forward to using the new templated test DB!

src/models/teams.rs
thesuzerain: Whoops- this is an artifact from that test I commented out.

src/models/teams.rs
```diff
@@ -79,10 +84,10 @@ impl ProjectPermissions {
 }
 
 bitflags::bitflags! {
-    #[derive(Copy, Clone, Debug)]
+    #[derive(Copy, Clone, Debug, PartialEq, Eq)]
     pub struct OrganizationPermissions: u64 {
         const EDIT_DETAILS = 1 << 0;
-        const EDIT_BODY = 1 << 1;
+        const UNUSED = 1 << 1; // a currently unused permission, formerly EDIT_BODY
```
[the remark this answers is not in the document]
thesuzerain: was trying to avod breakage in staging- but probably doesnt matter

Geometrically (approved):

PR #733, opened 2023-10-18
OAuth 2.0 Authorization Server [MOD-559]

OmegaJak:
Implements OAuth2.0 for the Modrinth backend, allowing Modrinth users to authorize third-party applications to act on their behalf.

This attempts to stick roughly to the [IETF RFC 6749](https://datatracker.ietf.org/doc/html/rfc6749#autoid-2) spec, but I'm sure there were a few deviations.

Here's a rough diagram of the flow:
d2 (1)

The beginning of the flow is a bit nonstandard as it routes first to the frontend, then to actual authorization server (labrinth), as we need the frontend to attach the appropriate header indicating a user is signed in to Modrinth.

This PR also tries to include the endpoints the frontend will need for users to create and manage their own OAuth apps, and to manage their authorized applications.


src/queue/session.rs
```diff
@@ -102,29 +113,51 @@ impl AuthQueue {
...
+            process_oauth_access_tokens(oauth_access_token_queue, &mut transaction).await?;
+
             transaction.commit().await?;
         }
 
         Ok(())
     }
 }
+
+async fn process_oauth_access_tokens(
```
Geometrically, line 145: why make this is a separate method? it doesn't seem to be used anywhere else

tests/files/dummy_data.sql
```diff
@@ -45,6 +45,25 @@ INSERT INTO categories (id, category, project_type) VALUES
...
+    )
+VALUES (
+        1,
+        'oauth_client_alpha',
+        NULL,
+        $1,
+        '4dbff86cc2ca1bae1e16468a05cb9881c97f1753bce3619034898faa1aabe429955a1bf8ec483d7421fe3c1646613a59ed5441fb0f321389f77f48a879c7b1f1',
+        3
+    );
+INSERT INTO oauth_client_redirect_uris (id, client_id, uri) VALUES (1, 1, 'https://modrinth.com/oauth_callback');
```
Geometrically, line 65: why is this included in the migration. it doesn't seem to be needed IMO, since you can create/generate this data via api routes, which should be preferred over raw database insertion

src/models/oauth_clients.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub id: OAuthRedirectUriId,
+    pub client_id: OAuthClientId,
+    pub uri: String,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct OAuthClientCreationResult {
+    pub client_secret: String,
+    pub client: OAuthClient,
+}
```
Geometrically, line 34: IMO you shouldn't have two separate structs. Just make client_secret an optional field on the client and set it to null by default

src/models/oauth_clients.rs
```diff
@@ -0,0 +1,100 @@
...
+#[derive(Deserialize, Serialize)]
+pub struct OAuthClientAuthorizationInfo {
+    pub id: OAuthClientAuthorizationId,
+    pub client_id: OAuthClientId,
+    pub user_id: UserId,
+    pub scopes: Scopes,
+    pub created: DateTime<Utc>,
+    pub client_name: String,
+    pub client_icon_url: Option<String>,
+    pub client_created_by: UserId,
```
Geometrically, line 61: imo these shouldn't be included in this struct. the consumer should just fetch this data separately

src/models/users.rs
```diff
@@ -89,6 +89,22 @@ impl From<DBUser> for User {
...
+        if self.role.is_mod() || self.id == client_owner_id {
+            Ok(())
+        } else {
+            Err(crate::routes::ApiError::CustomAuthentication(
+                "You don't have sufficient permissions to interact with this OAuth application"
+                    .to_string(),
+            ))
+        }
+    }
+}
```
Geometrically, line 106: weird place to put this method. imo this should go in the auth checks file or be a private func in the module that uses it

src/routes/v3/oauth_clients.rs
```diff
@@ -0,0 +1,387 @@
...
+pub fn config(cfg: &mut web::ServiceConfig) {
+    cfg.service(oauth_client_create);
+    cfg.service(oauth_client_edit);
+    cfg.service(oauth_client_delete);
+    cfg.service(get_user_clients);
+    cfg.service(get_user_oauth_authorizations);
+    cfg.service(revoke_oauth_authorization);
+}
+
+#[get("user/{user_id}/oauth_apps")]
```
Geometrically, line 45: this route should just be /oauth_apps IMO. it's more consistent with how we do PATs and sessions and notifs

src/routes/v3/oauth_clients.rs
```diff
@@ -0,0 +1,387 @@
...
+    let client_secret = generate_oauth_client_secret();
+    let client_secret_hash = DBOAuthClient::hash_secret(&client_secret);
+
+    let redirect_uris =
+        create_redirect_uris(&new_oauth_app.redirect_uris, client_id, &mut transaction).await?;
+
+    let client = OAuthClient {
+        id: client_id,
+        icon_url: new_oauth_app.icon_url.clone(),
+        max_scopes: new_oauth_app.max_scopes,
```
Geometrically, line 137: make sure you are checking that they are not requesting restricted scopes here (and anywhere else scope issuance is involved!)

src/routes/v3/oauth_clients.rs
```diff
@@ -0,0 +1,387 @@
...
+
+        transaction.commit().await?;
+
+        Ok(HttpResponse::Ok().body(""))
+    } else {
+        Ok(HttpResponse::NotFound().body(""))
+    }
+}
+
+#[get("user/oauth_authorizations")]
```
Geometrically, line 276: also think the user prefix should be removed from this

src/routes/v3/oauth_clients.rs
```diff
@@ -0,0 +1,387 @@
...
+    let authorizations =
+        OAuthClientAuthorization::get_all_for_user(current_user.id.into(), &**pool).await?;
+
+    let mapped: Vec<models::oauth_clients::OAuthClientAuthorizationInfo> =
+        authorizations.into_iter().map(|a| a.into()).collect_vec();
+
+    Ok(HttpResponse::Ok().json(mapped))
+}
+
+#[delete("user/oauth_authorizations/{client_id}")]
```
Geometrically, line 302: see above comment

src/routes/v3/oauth_clients.rs
```diff
@@ -0,0 +1,387 @@
...
+        .collect::<String>()
+}
+
+async fn get_oauth_client_from_str_id(
+    client_id: String,
+    exec: impl sqlx::Executor<'_, Database = sqlx::Postgres>,
+) -> Result<Option<OAuthClient>, ApiError> {
+    let client_id = models::ids::OAuthClientId::parse(&client_id)?.into();
+    Ok(OAuthClient::get(client_id, exec).await?)
+}
```
Geometrically, line 341: this is something that should be a part of the database func/api. see get string methods in database

src/routes/v3/oauth_clients.rs
```diff
@@ -0,0 +1,387 @@
...
+use crate::database::models::oauth_client_item::OAuthClient as DBOAuthClient;
+
+pub fn config(cfg: &mut web::ServiceConfig) {
+    cfg.service(oauth_client_create);
+    cfg.service(oauth_client_edit);
+    cfg.service(oauth_client_delete);
+    cfg.service(get_user_clients);
+    cfg.service(get_user_oauth_authorizations);
+    cfg.service(revoke_oauth_authorization);
+}
```
Geometrically, line 43: i think a better architecture would be having independent routes for retrieving oauth clients by ids, like single and multiple. and then other routes can just refer to the ids

Geometrically (changes requested):

src/auth/oauth/errors.rs
```diff
@@ -0,0 +1,167 @@
...
+        valid_redirect_uri: &ValidatedRedirectUri,
+    ) -> Self {
+        Self {
+            error_type: err.into(),
+            state: state.clone(),
+            valid_redirect_uri: Some(valid_redirect_uri.clone()),
+        }
+    }
+}
+
```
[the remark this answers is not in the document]
OmegaJak: This is to adhere to the OAuth spec as outlined in sections 4.1.2.1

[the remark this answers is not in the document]
OmegaJak: This is to adhere to oauth spec as outlined in [sections 4.1.2.1](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1) and [5.2](https://datatracker.ietf.org/doc/html/rfc6749#section-5.2).

Basically, instead of doing what we do for ApiError by returning error code + json-serialized error description, we need to do one of two things:
1. If the redirect uri is valid, then redirect the user to the callback location with error information provided as query parameters
2. If the redirect uri is invalid, generally, then we return an error directly
As we do these, to adhere to the spec, we need to use the codes and terms as outlined in RFC 6749.

Given this amount of separate behavior and the importance of isolating it from change that may come to ApiError that could bring it off-spec, it made sense to give it its own error type.

src/auth/oauth/mod.rs
```diff
@@ -0,0 +1,446 @@
...
+) -> Result<HttpResponse, OAuthError> {
+    let user = get_user_from_headers(
+        &req,
+        &**pool,
+        &redis,
+        &session_queue,
+        Some(&[Scopes::USER_AUTH_WRITE]),
+    )
+    .await
+    .map_err(OAuthError::error)?
```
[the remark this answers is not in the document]
OmegaJak: Yeah, not thrilled with how cumbersome the oauth error type turned out to be, but as outlined in my comment above, we need the error to take one of two forms - either a redirect with url-encoded error info or an error response directly back to the user agent.

So each time we're doing error handling in the oauth methods, we need to choose which of those is a better match - so we can't just let an auto-conversion happen using the `?` operator. It's critical to security that we don't redirect the user to the provided callback if the provided callback is invalid.

src/auth/oauth/mod.rs
```diff
@@ -0,0 +1,446 @@
...
+
+pub mod errors;
+pub mod uris;
+
+pub fn config(cfg: &mut ServiceConfig) {
+    cfg.service(
+        scope("auth/oauth")
+            .service(init_oauth)
+            .service(accept_client_scopes)
+            .service(reject_client_scopes)
```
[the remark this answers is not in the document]
OmegaJak: Yeah that's the idea - the user is asked to authorize in our own frontend and the 'yes' button goes to accept while the 'no' button goes to reject

src/database/models/oauth_client_authorization_item.rs
```diff
@@ -0,0 +1,142 @@
...
+pub struct OAuthClientAuthorizationWithClientInfo {
+    pub id: OAuthClientAuthorizationId,
+    pub client_id: OAuthClientId,
+    pub user_id: UserId,
+    pub scopes: Scopes,
+    pub created: DateTime<Utc>,
+    pub client_name: String,
+    pub client_icon_url: Option<String>,
+    pub client_created_by: UserId,
+}
```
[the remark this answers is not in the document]
OmegaJak: After our discussion in discord I thought this was the style we wanted, I must have misinterpreted. I will rework so the frontend gets that info with another request

src/queue/session.rs
OmegaJak: I think the code in this file would be better if the `index` method called a small number of individual methods for the different things it does - `set_last_login_data`, `remove_expired_sessions`, `update_pat_last_used`, and `updated_oauth_access_token_last_used`. Then you can read the top level method and understand what's happening quickly, while being able to drill down and focus on one of the bits of functionality if you need to.

Pulling this chunk into a method was a step towards that, without refactoring the whole file.

tests/files/dummy_data.sql
OmegaJak: Yeah totally agree, I wish I didn't have to. It's because of the client secret and how dummy data works now. We only run the code that interacts with the API to insert dummy data once per dummy data version - so most of the time when you run a test, that code inserting through the API isn't being run. Instead, we're just using methods to get data through the API. Since the client secret is generated by labrinth and isn't stored in its original form for security (we just keep the hash), it's on the creator of the oauth client to remember the secret.

Inserting directly into the database with a hash of a known, hardcoded secret was the best solution I saw to that problem.

src/queue/session.rs
OmegaJak: I'll be renaming this method to `updated_oauth_access_token_last_used`, which I think is better.

src/models/oauth_clients.rs
OmegaJak: I disagree and think that's a counterproductive pattern. One of the best things about Rust's expressive type system is that it can be used to make invalid states unrepresentable. This simplifies code, removes possible bugs, and makes it easier to read.

It would be invalid for a client secret to be returned at any time besides during creation. It would also be invalid for a client secret _not_ to be returned when we just created a client. By using the same struct everywhere, both of those invalid scenarios would become possible bugs and need to be guarded against. When they're separate types, those bugs are impossible, and developers don't have to waste any time considering how to handle them.

src/models/oauth_clients.rs
OmegaJak: Yep same misunderstanding, will rework to go through a different route

src/models/users.rs
OmegaJak: The method is on the struct that has the data it operates on. This makes it easier to discover that the method exists when you need it (helping to avoid duplication) and makes it less work to call.

The other bit of info it needs is the owner of the client - so another good place I could see this living is on the OAuthClient model. That might be better, might do that. Then it would also (kind of) be in the same conceptual module as where it's used.

src/routes/v3/oauth_clients.rs
OmegaJak: Sure, that makes sense. Do you think it should still take in {user_id} or should it just return the current user's clients based on the header, like pats?

src/routes/v3/oauth_clients.rs
OmegaJak: oo good call, that would have been a big miss. I'll do a pass over all the scopes usage and add handling for that.

src/routes/v3/oauth_clients.rs
OmegaJak: Will do

src/routes/v3/oauth_clients.rs
OmegaJak: Will do

src/routes/v3/oauth_clients.rs
OmegaJak: Oh yeah duh, thanks, will do

src/routes/v3/oauth_clients.rs
OmegaJak: Sure, will do

src/auth/oauth/errors.rs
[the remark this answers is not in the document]
thesuzerain: This makes sense- as oauth2 is going to involve other people using our API, to avoid people reporting issues in the future its worth following the existing specfiications IMO

src/auth/oauth/mod.rs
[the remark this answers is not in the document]
thesuzerain: could we set it up so that one of the two (the one that doesnt redirect to the provided callback) is ?-able? Or is this just ensuring we dont use ? no matter what the case so the programmer wont accidentally mix them up

src/routes/v3/oauth_clients.rs
thesuzerain: Depends on the anticipated use of it. If it was checking *other* user's apps (as a moderator, for instance) then {user_id} would make sense, but given that the code only accepts the user itself, I would say it doesnt need user_id

src/routes/v3/oauth_clients.rs
```diff
@@ -0,0 +1,387 @@
...
+
+    #[validate(
+        custom(function = "crate::util::validate::validate_url"),
+        length(max = 255)
+    )]
+    pub icon_url: Option<String>,
+
+    pub max_scopes: Scopes,
+
+    #[validate(length(min = 1))]
```
thesuzerain, line 98: does this need to be validated to 1? I dont see a big issue with a WIP oauth scope that simply doesnt allow usage yet

src/auth/oauth/mod.rs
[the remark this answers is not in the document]
OmegaJak: Hmm interesting yeah we could do that - basically have it default to OAuthError::error. That's the safer of the two options security-wise, so if we do that when we meant to redirect it's not the end of the world, the third party just won't get a callback telling them what happened. Plus, OAuthError::error is the one that needs no additional parameters, while OAuthError::redirect does.

I think there is some value in making sure we don't mix them up, but it's probably not worth the burden of mapping it all over the place. Thanks for the suggestion, I'll do that

src/routes/v3/oauth_clients.rs
OmegaJak: Are you referring to the `redirect_uris` here? I suppose we could drop the validation, but without any valid redirect URIs you won't be able to do any OAuth stuff. One of the first things it does when starting authorization is heck for a valid redirect uri.

Do you think it would be better to let the user create an 'invalid' application and discover that later when they try to authorize? That seems maybe worse to me but I'm not sure, what do you think?

thesuzerain: Yeah I am. 
I dont think this is a big deal either way just thinking out loud here. I definitely found use in incomplete WIP oauth stuff when I was doing Ory, but I could also see the value in making it harder for the user to make mistakes. Happy either way

src/models/pats.rs
```diff
@@ -103,6 +103,20 @@ bitflags::bitflags! {
...
+        // modify an OAuth client
+        const OAUTH_CLIENT_WRITE = 1 << 41;
+        // delete an OAuth client
+        const OAUTH_CLIENT_DELETE = 1 << 42;
+
+        // read a user's OAuth authorizations
+        const USER_OAUTH_AUTHORIZATIONS_READ = 1 << 43;
+        // write a user's OAuth authorizations
+        const USER_OAUTH_AUTHORIZATIONS_WRITE = 1 << 44;
+
```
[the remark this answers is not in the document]
OmegaJak: Alright I replaced all of the ones I added with that and pointed the methods I had using the new ones to it... I have no intuition about what the scopes should be so probably got it wrong, would appreciate review

src/auth/oauth/mod.rs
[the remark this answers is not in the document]
OmegaJak: Alright we're using `?` now for the error situations, still need to be explicit for redirects

src/database/models/oauth_client_authorization_item.rs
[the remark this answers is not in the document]
OmegaJak: Reworked

src/queue/session.rs
OmegaJak: Renamed

src/models/oauth_clients.rs
OmegaJak: Reworked

src/models/users.rs
OmegaJak: Moved into `auth/checks.rs` AND is accessible through OAuthClient, thanks to ✨Rust traits✨

src/routes/v3/oauth_clients.rs
OmegaJak: When creating a client we error if restricted scopes are requested, and when issuing a token we now sutract restricted scopes just to be sure.

src/routes/v3/oauth_clients.rs
OmegaJak: Alright, reworked the routes structure to the following:    
    Get user clients: GET /user/{user_id}/oauth_apps
        - Gets the clients for the current user based on header
        - Returns vec of client ids, no details
    Get client(s): GET /oauth/apps
	    - With json body { ids=[{id_1}, {id_2}, ...] }
    Get client: GET /oauth/app/{client_id}
    Create client: POST /oauth/app
    Edit client: PATCH /oauth/app/{client_id}
    Delete client: DELETE /oauth/app/{client_id}
    Get current user authorizations: GET /oauth/authorizations
        - Returns vec of user's authorizations with detail about authorization, but only client id, no other client details
    Revoke current user authorization: DELETE /oauth/authorizations?client_id={client_id}

src/routes/v3/oauth_clients.rs
OmegaJak: Yeah I'm torn too. @Geometrically you have an opinion here?

src/routes/v3/oauth_clients.rs
OmegaJak: Actually, on second thought - I don't think that should be the database code's concern. Parsing from strings/json is something we can let actix+serde do automatically using the `Deserialize` implementations we have for the ids. I've reworked things so external ids are automatically deserialized in this way and removed this method.

src/auth/checks.rs
```diff
@@ -8,6 +8,26 @@ use crate::routes::ApiError;
...
+}
+
+impl<'a, T, A> ValidateAllAuthorized for T
+where
+    T: IntoIterator<Item = &'a A>,
+    A: ValidateAuthorized + 'a,
+{
+    fn validate_all_authorized(self, user_option: Option<&User>) -> Result<(), ApiError> {
+        self.into_iter()
+            .map(|c| c.validate_authorized(user_option.clone()))
```
Geometrically: see clippy msg

src/auth/checks.rs
```diff
@@ -8,6 +8,26 @@ use crate::routes::ApiError;
...
+            .collect()
```
Geometrically: see clippy

src/models/oauth_clients.rs
Geometrically: Hmm, that's a fair point but I think you should use serde flatten in this case

src/routes/v3/oauth_clients.rs
Geometrically: Agree with wyatt it should allow to be empty and just error in the oauth flow

Geometrically (changes requested):

src/models/oauth_clients.rs
OmegaJak: Great idea, done! Love it.

src/routes/v3/oauth_clients.rs
OmegaJak: Alright cool, removed the validation, should be able to create a client without any redirect URIs now, and you'll still get a nice error message later telling you that that's why it's not working

Geometrically (approved):

PR #739, opened 2023-10-27
Feed system, user/organization following  [MOD-535]

OmegaJak:


src/database/models/project_item.rs
```diff
@@ -348,16 +348,6 @@ impl Project {
         if let Some(project) = project {
             Project::clear_cache(id, project.inner.slug, Some(true), redis).await?;
 
-            sqlx::query!(
```
OmegaJak, line 351: I just happened to notice this while reading this code - this is done again beneath the delete from mods_gallery. Am I missing something, or was this just unnecessary?

migrations/20231023212957_feeds.sql
```diff
@@ -0,0 +1,20 @@
...
+    triggerer_id_type text NULL,
+    event_type text NOT NULL,
+    metadata jsonb NULL,
+    created timestamptz NOT NULL DEFAULT CURRENT_TIMESTAMP
+);
+CREATE INDEX events_targets ON events (
+    target_id,
+    target_id_type,
+    event_type
+);
```
Geometrically, line 15: what are event targets for? imo we should query events on the route based on the user's follows. events should just be registered in the database to be queried later.

migrations/20231023212957_feeds.sql
```diff
@@ -0,0 +1,20 @@
...
+CREATE INDEX events_triggerers ON events (
```
Geometrically, line 16: same here, what are event triggerers?

src/database/models/event_item.rs
```diff
@@ -0,0 +1,285 @@
...
+use chrono::{DateTime, Utc};
+use itertools::Itertools;
+use sqlx::postgres::{PgHasArrayType, PgTypeInfo};
+use std::convert::{TryFrom, TryInto};
+
+#[derive(sqlx::Type, Clone, Copy)]
+#[sqlx(type_name = "text")]
+#[sqlx(rename_all = "snake_case")]
+pub enum EventType {
+    ProjectCreated,
```
Geometrically, line 14: version created/ project updated event is missing

src/database/models/creator_follows.rs
Geometrically, line 1: dislike having this in a separate file, this goes against the code practice we have so far. I also think macros make this a little unclear. 

IMO each following/unfollowing func should go in user_item and creator_item respectively. You can remove the get followers func bc we don't make followers of projects/users/orgs public right now

src/database/models/event_item.rs
```diff
@@ -0,0 +1,285 @@
...
+                event_type as "event_type: _",
+                metadata,
+                created
+            FROM events e
+            WHERE 
+                (target_id, target_id_type, event_type) 
+                = ANY(SELECT * FROM UNNEST ($1::bigint[], $2::text[], $3::text[]))
+            OR
+                (triggerer_id, triggerer_id_type, event_type) 
+                = ANY(SELECT * FROM UNNEST ($4::bigint[], $5::text[], $6::text[]))
```
Geometrically, line 197: i've seen you do this in a couple places, but i'm not sure why select unnest is needed here. you can just do ANY($1) and pass in a rust array and it works.

src/database/models/event_item.rs
```diff
@@ -0,0 +1,285 @@
...
+            &target_event_types[..] as &[EventType],
+            &triggerer_ids[..],
+            &triggerer_id_types[..] as &[IdType],
+            &triggerer_event_types[..] as &[EventType]
+        )
+        .fetch_all(exec)
+        .await?
+        .into_iter()
+        .map(|r| r.try_into())
+        .collect::<Result<Vec<_>, _>>()?)
```
Geometrically, line 211: see clippy warning

src/database/models/project_item.rs
Geometrically: yeah you can remove it if it's done twice

src/models/feed_item.rs
Geometrically, line 1: this should just be called "feed.rs". not feed_item"

src/routes/v2/project_creation.rs
```diff
@@ -748,11 +749,15 @@ async fn project_create_inner(
             }
         }
 
+        let organization_id = project_create_data.organization_id.map(|id| id.into());
+        insert_project_create_event(project_id, organization_id, &current_user, transaction)
+            .await?;
```
Geometrically, line 754: the project creation event should not happen here. Users should only be sent notifications if a project has been approved / has a public status. You can put this in the code where we send discord webhooks to the public channel (in project edit)

src/routes/v2/project_creation.rs
```diff
@@ -889,6 +894,27 @@ async fn project_create_inner(
     }
 }
 
+async fn insert_project_create_event(
```
Geometrically, line 897: i think this should be inlined since it's not used anywhere else

src/routes/v3/follow.rs
```diff
@@ -0,0 +1,107 @@
...
+    }
+}
+
+impl HasId<db_models::OrganizationId> for db_models::Organization {
+    fn id(&self) -> db_models::OrganizationId {
+        self.id
+    }
+}
+
+pub async fn follow<T, TId, Fut1, Fut2>(
```
Geometrically, line 29: this pattern also makes the code very hard to follow, I would just use the same pattern project following has, with 2 separate routes for both orgs and users, and put the code there. I understand it removes some repeated code, but it's not worth the cost esp bc nothing else will have following, this seems like you're designing a system with the intention we'll have 10+ things support following

src/routes/v3/users.rs
Geometrically, line 1: see my initial comment on feeds, will rev this file later since I imagine it might change later

Geometrically (changes requested):

migrations/20231023212957_feeds.sql
OmegaJak: Maybe I need better names... how I modeled it is that an event _triggerer_ causes an event to happen to an event _target_.  So a triggerer would be someone like a user, organization, maybe an automatic process - whatever, it's the thing that caused the event to happen. Meanwhile a target is the thing the event happens to - for project-related events, it's the project. If we want to make it so someone can follow, idk, something like changes to organizations, the organization itself would be the target and the triggerer would be the user that initiated the change. The goal with this modeling was to reflect the fact that you might want to follow one (or both) sides of an event - you may be interested in what a user/organization is doing, or maybe you're just interested in one specific project.

I suppose we could drop the concept of a triggerer and insert two events for project creation - one event tied to the id of the user/org that did it and another tied to the id of the project or whatever, but that seems worse to me.

As for this specific line, it's just creating an index that matches how we query the table in an attempt to improve query performance.

migrations/20231023212957_feeds.sql
OmegaJak: See above, I tried to explain the two concepts. Like the other, this is just an index.

src/database/models/event_item.rs
OmegaJak: I wasn't under the impression we wanted those implemented in this first pass and assumed we would follow up this PR with ones implementing those.

Would you like me to implement those in this PR?

src/database/models/creator_follows.rs
OmegaJak: My understanding was that the pattern is to generally have a file dedicated to interacting with one table/conceptual 'thing' in the database (and maybe its supporting tables). `user_follows` and `organization_follows` are their own tables and are conceptually separate to me from users/orgs, so I'm not sure I see the benefit of putting the code that interacts with them into `user_item.rs` and `organization_item.rs`.

I agree, the macro has a cost, but given that these tables are almost identical in the data they store and are identical in how we interact with them (to the point that I heavily considered putting them into the same table, but didn't only to maintain foreign key linkages), I couldn't stomach duplicating all of their DB code. I wish I could do it with traits, but, afaik, we have to use macros to use the statically validated sqlx macros.

Will remove the unused method.

src/database/models/event_item.rs
OmegaJak: Not sure how that would work. This is the only way I could find to get the semantics I wanted. This has a different meaning from doing something like 
```sql
WHERE triggerer_id = ANY($1) AND triggerer_id_type = ANY($2) AND event_type = ANY($3)
```

We need to match specific combination triples of these values, not just any that exist in any of the three arrays. I don't know how to do that besides the way I did. _Maybe_ there's a way using sqlx type derives, but I don't think you can do that for anonymous composite types - you need a named composite type in the schema. If you know how, could you give me an example?

src/database/models/event_item.rs
OmegaJak: Fixed this and others, the two remaining clippy warnings are in code that's going to change.

src/routes/v3/follow.rs
OmegaJak: Yeah, this turned out much clunkier than I wanted. I disagree with the implication that something needs 10+ usages to be worth avoiding duplication, but in this case, I think the benefit is slim enough that I'll inline and duplicate them as you'd like.

This could have been done much more elegantly with traits, but unfortunately `async fn` isn't yet valid in trait definitions, so this was the best I could come up with. That should be coming with [Rust 1.75](https://releases.rs/docs/1.75.0/) in December, too bad we don't have it now.

src/routes/v2/project_creation.rs
OmegaJak: I disagree because I think it improves the readability. If you don't care about the event insertion while you're reading through `project_create_inner`, you can see the name and move on, giving it no further thought. If you do care, it has a descriptive name telling you what it does and is a simple F12/`View Declaration` away and then you get to focus just on the details of how we generate the event, without the distractions of the all the other things we're doing in that method.

I do agree that the benefit here is relatively slim because it ended up being a pretty short method, but I still think it's easier to read. If you'd really like, though, I'll inline it.

src/models/feed_item.rs
OmegaJak: Renamed to `feeds.rs`

src/routes/v2/project_creation.rs
Geometrically: Yeah that seems fine

src/database/models/event_item.rs
Geometrically: Yeah I think it would be worth implementing

migrations/20231023212957_feeds.sql
Geometrically: ah, that makes sense, I mistook this for a separate table

src/database/models/event_item.rs
Geometrically: ohh, I see, that makes sense

src/routes/v3/follow.rs
Geometrically: I was exaggerating yeah but I think it would be worth inlining these

src/routes/v2/project_creation.rs
Geometrically: See disc

src/database/models/creator_follows.rs
```diff
@@ -0,0 +1,97 @@
...
+
+macro_rules! impl_follow {
+    ($target_struct:ident, $table_name:tt, $target_id_type:ident, $target_id_ctor:expr) => {
+        impl $target_struct {
+            pub async fn insert(
+                &self,
+                exec: impl sqlx::Executor<'_, Database = sqlx::Postgres>,
+            ) -> Result<(), DatabaseError> {
+                sqlx::query!(
+                    " INSERT INTO " + $table_name + " (follower_id, target_id) VALUES ($1, $2)",
```
Geometrically, line 47: does sqlx support using the prepared statement for the table name? like INSERT INTO $1? I think it would be worth trying bc I am uneasy abt string concat in sql statements (this is totally safe, but I'm worried it's a bad practice)

src/routes/v3/follow.rs
OmegaJak: Done

src/database/models/creator_follows.rs
OmegaJak: As far as I can tell, no, unfortunately (https://stackoverflow.com/questions/25631006/postgres-sqlx-prepared-statement-with-table-name-bindvar, links https://www.postgresql.org/docs/9.1/xfunc-sql.html). It's a Postgres limitation. Not a huge fan either, but the only other alternatives I see are duplicating or using dynamic (unchecked) queries, which would still require string concatenation (but not the macro, at least).

src/routes/v2/projects.rs
```diff
@@ -526,6 +529,21 @@ pub async fn project_edit(
...
+                            .await?;
+                    if let Some(owner_id) = owner_id {
+                        insert_project_publish_event(
+                            id.into(),
+                            project_item.inner.organization_id,
+                            owner_id,
+                            &mut transaction,
+                        )
+                        .await?;
+                    }
```
Geometrically, line 546: hmm, I think this should use the org ID if the project belongs to the org, and if not it should use the owner ID. could it be changed to that?

Geometrically (changes requested):

PR #740, opened 2023-10-28
Version ordering [MOD-551]

OmegaJak:


Geometrically (approved):

src/routes/v2/versions.rs
Geometrically, line 1: make sure to incorporate other routes that rely on ordering, like updating, version slug route, maven, etc

Geometrically (changes requested):

src/routes/v2/versions.rs
OmegaJak: Alright, implemented for maven and every other bit of code I could find that was previously sorting by `date_published` in Rust (as opposed to in SQL).

Geometrically (approved):

PR #746, opened 2023-11-06
Fix download counts

Geometrically:


src/routes/mod.rs
```diff
@@ -176,3 +176,5 @@ impl actix_web::ResponseError for ApiError {
         })
     }
 }
+
```
thesuzerain, line 179: Fine for now as we need to get this into prod but this seems like feels like there should be a way to structure the these changes so that we don't need to do this.

thesuzerain (approved):

PR #752, opened 2023-11-10
Knossos Oauth 2 Flow Changes

darling:


src/auth/oauth/errors.rs
```diff
@@ -63,7 +63,7 @@ impl actix_web::ResponseError for OAuthError {
             | OAuthErrorType::ScopesTooBroad
             | OAuthErrorType::AccessDenied => {
                 if self.valid_redirect_uri.is_some() {
-                    StatusCode::FOUND
+                    StatusCode::OK
```
thesuzerain, line 66: What is the issue that is causing this to be needed? I remember Jackson was being particular about the error codes, is this fixing it to adhere to the oauth spec or was oauth spec not working with knossos?

src/auth/oauth/mod.rs
```diff
@@ -413,9 +415,9 @@ async fn init_oauth_code_flow(
     let redirect_uri = append_params_to_uri(&redirect_uris.validated.0, &redirect_params);
 
     // IETF RFC 6749 Section 4.1.2 (https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2)
-    Ok(HttpResponse::Found()
-        .append_header((LOCATION, redirect_uri))
-        .finish())
+    Ok(HttpResponse::Ok()
```
thesuzerain, line 418: See above
the spec seems to say should be a 302- if there is an issue with it for our setup I would recommend leaving a comment here saying this deviates from the spec (we shouldnt have the spec as a comment right above a place it deviates from it)

src/auth/oauth/errors.rs
```diff
@@ -94,10 +94,9 @@ impl actix_web::ResponseError for OAuthError {
                 redirect_uri = format!("{}&state={}", redirect_uri, state);
             }
 
-            redirect_uri = urlencoding::encode(&redirect_uri).to_string();
-            HttpResponse::Found()
-                .append_header(("Location".to_string(), redirect_uri))
-                .finish()
+            HttpResponse::Ok()
```
thesuzerain, line 97: See above

src/models/pats.rs
```diff
@@ -132,7 +132,10 @@ impl Scopes {
     }
 
     pub fn parse_from_oauth_scopes(scopes: &str) -> Result<Scopes, bitflags::parser::ParseError> {
-        let scopes = scopes.replace(' ', "|").replace("%20", "|");
+        let scopes = scopes
```
thesuzerain, line 135: Clippy issue here- make sure you run `cargo clippy` (and maybe `--fix`) :)

src/routes/v3/oauth_clients.rs
```diff
@@ -111,13 +112,19 @@ pub async fn get_client(
...
     pool: web::Data<PgPool>,
     redis: web::Data<RedisPool>,
     session_queue: web::Data<AuthQueue>,
 ) -> Result<HttpResponse, ApiError> {
-    let clients =
-        get_clients_inner(&info.into_inner().ids, req, pool, redis, session_queue).await?;
+    let ids: Vec<_> = info
+        .ids
+        .iter()
+        .map(|id| parse_base62(&id).map(ApiOAuthClientId))
```
thesuzerain, line 123: Clippy issue here as well- you dont need to dereference with `&id` it seems

thesuzerain (changes requested):

src/models/pats.rs
darling: Man clippy and rust in general is so cool

Geometrically (approved):

PR #754, opened 2023-11-12
Tests 3 restructure

thesuzerain:
- Restructures test to use V3 API (and adds V3 API functions)
- Makes tests pass, fixes issues in v3 route setup that caused tests to fail
- Added extra tests for permissions


migrations/20241110010322_adds_game_version_minmax.sql
```diff
@@ -0,0 +1,2 @@
+UPDATE loader_fields SET min_val = 1 WHERE field = 'game_versions';
+UPDATE loader_fields SET max_val = 20 WHERE field = 'game_versions';
```
Geometrically, line 2: why 20? before we didn't have a limit for game versions, why add one now?

src/auth/flows.rs
```diff
@@ -53,7 +55,9 @@ pub fn config(cfg: &mut ServiceConfig) {
             .service(set_email)
             .service(verify_email)
             .service(subscribe_newsletter)
-            .service(link_trolley),
+            .service(link_trolley)
+            // auth/oauth
+            .configure(oauth::config),
```
Geometrically, line 60: it's weird that this is under auth IMO, cld we just make it top level?

src/routes/v2/mod.rs
```diff
@@ -26,9 +26,14 @@ pub fn config(cfg: &mut actix_web::web::ServiceConfig) {
...
             .configure(admin::config)
             .configure(analytics_get::config)
+            // THESE NEED TO BE SEPARATED
+            // TODO
+            // TODO PANIC UNWRAP
             .configure(crate::auth::session::config)
             .configure(crate::auth::flows::config)
             .configure(crate::auth::pats::config)
+            // TODO ^
+            // need to be able to v2 these
```
Geometrically, line 36: seems to be todo here?

src/routes/v2/version_creation.rs
```diff
@@ -95,6 +95,10 @@ pub async fn version_create(
                 json!(legacy_create.game_versions),
             );
 
+            // TODO: This needs to actually get the SideType from somewhere- another version? Default?
+            fields.insert("client_side".to_string(), json!("required"));
+            fields.insert("server_side".to_string(), json!("optional"));
+
```
Geometrically, line 101: hmm, I think it should just default to the prev version. i would remove this though in this pr bc there is a later linear issue for redoing the side type system

Geometrically (changes requested):

src/routes/v2/mod.rs
thesuzerain: I pushed this out to another PR, but forgot to remove my... aggressive TODO comments. Will reword the comment but intend to do in another PR

migrations/20241110010322_adds_game_version_minmax.sql
thesuzerain: Ah fair enough.

src/auth/flows.rs
thesuzerain: yup

src/routes/v2/version_creation.rs
thesuzerain: Sounds good. The reason i have it in is to keep tests passing rn- as it needs some kind of value in the conversion to call t he route successfully atm. I'll delete them when I merge the side types PR.

Geometrically (approved):

PR #763, opened 2023-11-22
Links

thesuzerain:
Fixes MOD-558


src/database/models/categories.rs
```diff
@@ -29,10 +29,10 @@ pub struct ReportType {
 }
 
 #[derive(Serialize, Deserialize)]
-pub struct DonationPlatform {
-    pub id: DonationPlatformId,
-    pub short: String,
+pub struct LinkPlatform {
+    pub id: LinkPlatformId,
     pub name: String,
+    pub donation: bool,
```
Geometrically, line 35: is this bool needed? I assume it's for legacy compat but I feel like there's a better way to do this

Geometrically: Like maybe we could hardcode? or do you think that's worse

src/routes/v2/projects.rs
```diff
@@ -333,18 +334,79 @@ pub async fn project_edit(
...
+            new_links.insert("discord".to_string(), None);
+        }
+    }
+
+    // In v2, setting donation links resets all other donation links
+    // (resetting to the new ones)
+    if let Some(donation_urls) = v2_new_project.donation_urls {
+        // Fetch current donation links from project so we know what to delete
+        // Todo ensure secure
+        let fetched_example_project = project_item::Project::get(&info.0, &**pool, &redis).await?;
```
Geometrically, line 375: unfinished todo?

Geometrically (changes requested):

src/database/models/categories.rs
thesuzerain: I kinda feel like this is the structurally to hardcoding but without having to actually hardcode it
It might also be useful for the frontend at some point to know which links are donation links without hardcoding (ie: 'see donation links' or w/e depending on how it 
I'm for it being this way, but dont feel that strongly about it so can change it if you prefer

Geometrically (approved):

Geometrically (approved):

PR #765, opened 2023-11-22
Payouts code

Geometrically:


migrations/20231114175920_new-payment-methods.sql
```diff
@@ -0,0 +1,28 @@
...
+ALTER TABLE payouts
+    ADD COLUMN method text NULL,
+    ADD COLUMN method_address text NULL,
+    ADD COLUMN platform_id text NULL,
+    ADD COLUMN fee numeric(40, 20) NULL,
+    ALTER COLUMN id TYPE bigint,
+    ALTER COLUMN id DROP DEFAULT;
+
+UPDATE payouts
+SET status = 'success';
```
thesuzerain, line 26: is this intended to hit every past payout?

src/models/v3/users.rs
```diff
@@ -59,11 +59,14 @@ pub struct User {
     pub github_id: Option<u64>,
 }
 
+// TODO: make v2 supported version of this
```
thesuzerain, line 62: TODO- for a different PR?

src/routes/v3/payouts.rs
```diff
@@ -0,0 +1,732 @@
...
+        )
+        .await?;
+
+    if &webhook_res.verification_status != "SUCCESS" {
+        return Err(ApiError::InvalidInput(
+            "Invalid webhook signature".to_string(),
+        ));
+    }
+
+    println!("{}", webhook_res.verification_status);
```
thesuzerain, line 96: println

src/routes/v3/payouts.rs
```diff
@@ -0,0 +1,732 @@
...
+pub async fn create_payout(
+    req: HttpRequest,
+    pool: web::Data<PgPool>,
+    redis: web::Data<RedisPool>,
+    body: web::Json<Withdrawal>,
+    session_queue: web::Data<AuthQueue>,
+    payouts_queue: web::Data<PayoutsQueue>,
+) -> Result<HttpResponse, ApiError> {
+    let (scopes, user) =
+        get_user_record_from_bearer_token(&req, None, &**pool, &redis, &session_queue)
```
thesuzerain, line 360: Is there a reason this uses a separate get user to other routes ? (get_user_from_headers or w/e)

src/routes/v3/payouts.rs
```diff
@@ -0,0 +1,732 @@
...
+    if !scopes.contains(Scopes::PAYOUTS_WRITE) {
+        return Err(ApiError::Authentication(
+            AuthenticationError::InvalidCredentials,
+        ));
+    }
+
+    let mtx = payouts_queue.lock_user_payouts(user.id.into());
+    let _guard = mtx.lock().await;
+
+    if user.balance < body.amount || body.amount < Decimal::ZERO {
```
thesuzerain, line 373: I dont think `body.amount < Decimal::ZERO` should return this error - maybe make it use the 'iwth draw more to cover the fee' one below?

thesuzerain (changes requested):

src/auth/flows.rs
```diff
@@ -623,6 +684,7 @@ impl AuthProvider {
                     avatar_url: gitlab_user.avatar_url,
                     bio: gitlab_user.bio,
                     name: gitlab_user.name,
+                    country: None,
```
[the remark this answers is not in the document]
thesuzerain: I also noticed that its saved in the db under paypal_country

migrations/20231114175920_new-payment-methods.sql
Geometrically: Yes by default the backend assumes them as successes

src/routes/v3/payouts.rs
Geometrically: yes it needs the raw paypal_id which aren't exposed in the public func

src/routes/v3/payouts.rs
Geometrically: If the body amount is negative it's invalid, this catches this

src/models/v3/users.rs
Geometrically: removed it since don't plan on doing it (this route is only used by us)

src/auth/flows.rs
```diff
@@ -707,11 +770,54 @@ impl AuthProvider {
...
+                        country: None,
                     }
                 } else {
                     return Err(AuthenticationError::InvalidCredentials);
                 }
             }
+            AuthProvider::PayPal => {
+                #[derive(Deserialize, Debug)]
+                pub struct PayPalUser {
+                    pub payer_id: String,
```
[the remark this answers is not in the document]
Geometrically: idk paypal's api just names their user id weird

src/auth/flows.rs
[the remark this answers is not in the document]
Geometrically: it's only meant for paypal, so I think the naming is fine here

migrations/20230919183129_trolley.sql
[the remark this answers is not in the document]
Geometrically: I can manually fix it, the old migration causes data loss so had to delete it

thesuzerain (approved):

PR #781, opened 2023-12-05
Optimize analytics queries

Geometrically:


thesuzerain (changes requested):
Looks good- just a couple clippy things.

thesuzerain (approved):

PR #796, opened 2023-12-13
Organization ownership

thesuzerain:
Organiztion owner structure changed as discussed.
- is_owner is unique
- projects owned by an org do not have a owner
- projects owned by an org do not have that orgs owner as a member (so permissions cant be overwritten)
- removing a project from an org you must specify to whom ownership of it goes
- when deleting an org, all projects become owned by the org owner.

Review carefully as it deals with ownership /permissions.


src/routes/v3/organizations.rs
```diff
@@ -522,7 +523,105 @@ pub async fn organization_delete(
...
+        SET is_owner = FALSE
+        WHERE tm.team_id = ANY($1)
+        ",
+        &organization_project_teams
+            .iter()
+            .map(|x| x.0 as i64)
+            .collect::<Vec<_>>()
+    )
+    .execute(&mut *transaction)
+    .await?;
```
Geometrically, line 606: this should be removed since it's an illegal state

src/routes/v3/organizations.rs
```diff
@@ -522,7 +523,105 @@ pub async fn organization_delete(
...
+
+    for project_team_without_owner in project_teams_without_owner {
+        let new_id =
+            crate::database::models::ids::generate_team_member_id(&mut transaction).await?;
+        let member = TeamMember {
+            id: new_id,
+            team_id: project_team_without_owner,
+            user_id: owner_id,
+            role: "Inherited Owner".to_string(),
+            is_owner: false,
```
Geometrically, line 583: set is_owner here

Geometrically: to true

src/routes/v3/organizations.rs
```diff
@@ -522,7 +523,105 @@ pub async fn organization_delete(
...
+        ",
+        &organization_project_teams
+            .iter()
+            .map(|x| x.0 as i64)
+            .collect::<Vec<_>>(),
+        owner_id.0
+    )
+    .execute(&mut *transaction)
+    .await?;
+
```
Geometrically, line 623: remove this (also make it so the owner of an org can never be a team member of the project

src/routes/v3/organizations.rs
```diff
@@ -622,6 +725,23 @@ pub async fn organization_projects_add(
...
+                "
+                UPDATE team_members
+                SET is_owner = FALSE
+                WHERE id = $1
+                ",
+                team_member_id as database::models::ids::TeamMemberId
+            )
+            .execute(&mut *transaction)
+            .await?;
+        }
```
Geometrically, line 743: also shouldn't be needed

src/routes/v3/organizations.rs
```diff
@@ -696,7 +824,78 @@ pub async fn organization_projects_remove(
...
+            organization.id,
+            data.new_owner.into(),
+            &**pool,
+        )
+        .await?
+        .ok_or_else(|| {
+            ApiError::InvalidInput(
+                "The specified user is not a member of this organization!".to_string(),
+            )
+        })?;
```
Geometrically, line 838: the user should be allowed to move it to a user who is part of the project's team but not the org's team as well

src/routes/v3/organizations.rs
```diff
@@ -696,7 +824,78 @@ pub async fn organization_projects_remove(
...
+        sqlx::query!(
+            "
+            UPDATE team_members
+            SET is_owner = TRUE
+            WHERE (id = $1)
+            ",
+            new_owner.id as database::models::ids::TeamMemberId
+        )
+        .execute(&mut *transaction)
+        .await?;
```
Geometrically, line 897: also see above, since it's redundant we can probs get rid of it

src/routes/v3/teams.rs
```diff
@@ -723,8 +724,14 @@ pub async fn transfer_ownership(
     .await?
     .1;
 
+    println!(
+        "Transferring owner from {} to {}",
+        current_user.id.0, new_owner.user_id.0
+    );
+    println!("Trasnfering team: {}", id.0);
```
Geometrically, line 731: print statements

Geometrically (changes requested):

src/routes/v3/organizations.rs
thesuzerain: and also there shouldnt be an owner of a project if there is an organization

Geometrically (approved):

PR #805, opened 2023-12-19
Misc testing improvements

thesuzerain:
- Scope tests now fully test v2 as well
- claned up dummy data a little bit
- Other hardcoded test functions now use testing api
- Asserts to a StatusCode now print response body on failure and all relevant asserts use that function


src/routes/v3/users.rs
```diff
@@ -93,6 +93,7 @@ pub async fn user_auth_get(
     redis: web::Data<RedisPool>,
     session_queue: web::Data<AuthQueue>,
 ) -> Result<HttpResponse, ApiError> {
+    println!("AUTH GET!");
```
Geometrically, line 96: print statements here

Geometrically (changes requested):

Geometrically (approved):

PR #813, opened 2023-12-27
Fixes failing tests

thesuzerain:
- Improves 500 errors happening randomly
- Disables a field on a test due to a known shortcoming causing random test failure
- Fixes the random search test data races by entirely separating the meilisearch indices amongst different tests


src/file_hosting/mock.rs
```diff
@@ -47,8 +47,9 @@ impl FileHost for MockHost {
     ) -> Result<DeleteFileData, FileHostingError> {
         let path = std::path::Path::new(&dotenvy::var("MOCK_FILE_PATH").unwrap())
             .join(file_name.replace("../", ""));
-        std::fs::remove_file(path)?;
-
+        if path.exists() {
+            std::fs::remove_file(path)?;
+        }
```
thesuzerain, line 52: I'm fairly sure this fixes the random test failing. It happens like 1/30 times inconsistently and as far as I can tell was traceable to this line- so I think the same test file is attempted to be deleted twice, and some odd file system data race is happening.

src/routes/v3/users.rs
```diff
@@ -65,23 +65,12 @@ pub async fn projects_list(
...
-            crate::database::Project::get_many_ids(&project_data, &**pool, &redis)
-                .await?
-                .into_iter()
-                .filter(|x| can_view_private || x.inner.status.is_searchable())
-                .map(Project::from)
-                .collect();
-
-        Ok(HttpResponse::Ok().json(response))
+        let projects: Vec<_> =
+            crate::database::Project::get_many_ids(&project_data, &**pool, &redis).await?;
```
thesuzerain, line 71: this change ported over from extracted_versions, makes the user route consistent with the other routes that only let you see some projects, and allosw the filter visible projeccts test down below to pass

Geometrically (approved):

PR #815, opened 2023-12-29
Shared modpacks routing

thesuzerain:
NOTE: **_DO NOT MERGE_** until changes noted in the migrations file are reverted.

Fixes MOD-576

Notable deviations from spec:
- Link generated is not immediately returned to the user, but attached to another route (so  can be called multiple times for multiple share links to be generated as they run out of uses/expire)
- Link generated is an API link to the join-profile route (with the appropriate url identifier). This will be changed to be a `modrinth://` link as we move from testing this with cloudflare to having the theseus implementation.


src/database/models/client_profile_item.rs
```diff
@@ -0,0 +1,518 @@
...
+        game_id: GameId,
+        game_name: String,
+        game_version_id: LoaderFieldEnumValueId,
+        game_version: String,
+    },
+    Unknown {
+        game_id: GameId,
+        game_name: String,
+    },
+}
```
Geometrically, line 52: Do we need an enum here? Since we already have games dynamically in the database. Couldn't we just add a jsonb metadata field with game metadata (and then remove game_id and game_name fields

Geometrically: Alternatively we could build this in with the loader fields system... but I don't think it's the best idea bc stuff like loader versions aren't on labrinth

Geometrically (changes requested):
- make the routes internal also

migrations/20231226012200_shared_modpacks.sql
```diff
@@ -0,0 +1,45 @@
...
+    icon_url varchar(255),
+    color integer NULL,
+    updated timestamptz NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    created timestamptz NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
+    loader_id int NOT NULL REFERENCES loaders(id), 
+    metadata jsonb NOT NULL DEFAULT '{}'::jsonb,
+
+    game_id int NOT NULL REFERENCES games(id),
+    game_version_id int NULL REFERENCES loader_field_enum_values(id) -- Minecraft java
```
Geometrically, line 14: why does this exist if the game version is also in the metadata?

src/models/v3/client/profile.rs
```diff
@@ -0,0 +1,146 @@
...
+
+        Self {
+            id: link_id,
+            url,
+            profile_id,
+            created: link.created,
+            expires: link.expires,
+        }
+    }
+}
```
Geometrically, line 146: The URL shouldn't be returned / parsed here. The frontend will generate it based on the link ID

src/models/v3/client/profile.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    /// Game-specific information
+    #[serde(flatten)]
+    pub game: ClientProfileMetadata,
+
+    /// Modrinth-associated versions
+    pub versions: Vec<VersionId>,
+    /// Overrides for this profile- only install paths are given,
+    /// hashes are looked up in the CDN by the client
+    pub override_install_paths: Vec<PathBuf>,
```
Geometrically, line 55: What does this mean? IMO we shouldn't have the versions or any install info be returned here. It should be a separate route that is only called when the pack is being installed. And it can only be accessed once a user has accepted the invite.

src/routes/internal/client/profiles.rs
```diff
@@ -0,0 +1,1292 @@
...
+}
+
+// Delete a client profile
+pub async fn profile_delete(
+    req: HttpRequest,
+    info: web::Path<(String,)>,
+    pool: web::Data<PgPool>,
+    redis: web::Data<RedisPool>,
+    session_queue: web::Data<AuthQueue>,
+) -> Result<HttpResponse, ApiError> {
```
Geometrically, line 502: This should delete override files from the CDN as long as they are unused by other shared packs

migrations/20231226012200_shared_modpacks.sql
```diff
@@ -0,0 +1,45 @@
...
+);  
+
+CREATE TABLE shared_profiles_mods (
+    shared_profile_id bigint NOT NULL REFERENCES shared_profiles(id),
+
+    -- for versions we have hosted
+    version_id bigint NULL REFERENCES versions(id), -- for versions
+
+    -- for cdn links to files we host directly
+    file_hash varchar(255) NULL,
```
Geometrically, line 25: We should restructure this. Override files should be a separate table which this table refers too IMO. And then it should refer to this here (we could just re-use the files table TBH).

Like other API objects, this would allow us to have SHA1 and SHA512 hashes. Shared modpacks will also share files a ton (ex: if someone has the same exact config file between packs, we shouldn't haave to host it twice or have duplicate database entries)

Geometrically: I think just using the files table would probably work well enough. It would also allow it so if someone did ever upload their mod to Modrinth, we could "re-link" it to the project dynamically and change it from an override to a Modrinth version ID. If that makes sense

src/routes/internal/client/profiles.rs
```diff
@@ -0,0 +1,1292 @@
...
+                web::scope("profile")
+                    .route("{id}", web::get().to(profile_get))
+                    .route("{id}", web::patch().to(profile_edit))
+                    .route("{id}", web::delete().to(profile_delete))
+                    .route("{id}/override", web::post().to(client_profile_add_override))
+                    .route(
+                        "{id}/override",
+                        web::delete().to(client_profile_remove_overrides),
+                    )
+                    .route("{id}/share", web::get().to(profile_share))
```
Geometrically, line 52: IMO this route is kind of pointless. We should just return share links in the main profile body. When would you get a profile and not get its share links?

src/routes/internal/client/profiles.rs
```diff
@@ -0,0 +1,1292 @@
...
+                    .route(
+                        "{id}/share/{url_identifier}",
+                        web::get().to(profile_link_get),
+                    )
+                    .route(
+                        "{id}/accept/{url_identifier}",
+                        web::post().to(accept_share_link),
+                    )
```
Geometrically, line 60: These routes are structured wrong. A share link would be like https://modrinth.com/share/aaJJJwwa. You won't know the profile ID. There should be a route to just get a profile from share link and accept a share link

Geometrically (changes requested):

src/models/v3/client/profile.rs
thesuzerain: Im nott sure I agree- I think it could be useful to be able to see the stuff youre going to install on your computer before accepting, or what's 'in/on' the shared modpack.

All this displays is what the versions are and the 'names' of the overrides.

PR #818, opened 2023-12-30
Remove mcmod.info validation for legacy forge versions

DJtheRedstoner:
Coremods/tweaker only mods do not need to contain this file. This is similar to a change for modern forge made in a5195920fa.


DJtheRedstoner:
Fixed by b596b1ca5298b5080675f5719ededd54b9296415

PR #845, opened 2024-01-09
Fix melisearch name replacements with operators other than :

JulianVennen:
It appears that some field names were changed in meilisearch and to keep api compatibility the old names are mapped to the new ones.

Unfortunately, this mapping only works when using the equality operator ':', not when using '=', '<', '<=', ...
This PR fixes that and supports all operators mentioned in the API documentation.

I also noticed that since split, not split_once was used during this replacement, part of the value would disappear if it contained another colon. This issue no longer exists in my fixed version.

Splitting the facet declaration into its parts is done by the method parse_facet, which I had to implement myself because split and split_once do not support passing a list of strings (only chars) as delimiters.

### DIFF
--- a/src/routes/v2/projects.rs
+++ b/src/routes/v2/projects.rs
@@ -93,7 +93,7 @@ pub async fn project_search(
                             facets
                                 .into_iter()
                                 .map(|facet| {
-                                    if let Some((key, operator, val)) = parse_facet(facet) {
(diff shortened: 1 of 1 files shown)


JulianVennen:
It appears that GitHub is having some issues which is causing the actions to break right now

thesuzerain:
LGTM- rerunning GH actions

Geometrically (approved):

PR #853, opened 2024-01-13
Index swapping when meilisearch reindex

thesuzerain:
Implements the following behaviour
- multiple indices exist at any given time
- When we re-index meilisearch, we create and populate a new index, while leaving the last one (to be searched, etc)
- When we are done indexing, we swap indices and delete the previous one


Geometrically (changes requested):
Looks good but you should keep the document deletion funcs

Geometrically (approved):

PR #863, opened 2024-01-16
Fix search query parameter validation

triphora:


src/search/mod.rs
```diff
@@ -191,7 +191,7 @@ pub async fn search_for_project(
 
     let offset: usize = info.offset.as_deref().unwrap_or("0").parse()?;
     let index = info.index.as_deref().unwrap_or("relevance");
-    let limit = info.limit.as_deref().unwrap_or("10").parse()?;
+    let limit = info.limit.as_deref().unwrap_or("10").parse::<usize>()?.min(100);
```
tizu69, line 194: shouldn't this be max instead of min?

triphora: No, `min` returns the minimum of the two. If the inputs are 10 and 100, it returns 10. If the inputs are 400 and 100, it returns 100.

tizu69 (approved):
aight lgtm then

Geometrically (approved):

PR #872, opened 2024-01-27
Fix #868 - allow Basic and Bearer auth 

### LINKED ISSUE #868: OAuth 2.0 behavior deviates from standard practice
### Describe the bug

The current implementation of the OAuth 2.0 flow deviates in certain key areas making it hard to interop with generic clients. I encountered these while using my own generic OAuth handler.

Most notably,
- Client authentication ([RFC 6749 §2.3.](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3)) is done using a non-standard authentication system where the client secret is sent "as-is" as the authentication header. The general standard is to use HTTP Basic authentication, which is mentioned in [§2.3.1.](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1).
- The API returns a Bearer token ([RFC 6750](https://datatracker.ietf.org/doc/html/rfc6750)), but does not support `Bearer <token>` authorization format despite being the format defined in [RFC 6750 §2.1.](https://datatracker.ietf.org/doc/html/rfc6750#section-2.1). Only the prefix-less token is supported.

### Steps to reproduce

1. Create an OAuth application
2. Initiate an authotization flow
3. Request the bearer access token (with `Authorization: Basic base64(client_id:client_secret)`)
4. Request the authenticated user data (with `Authorization: Bearer access_token`)

### Expected behavior

The flow to succeed and to get information about the authenticated user.

### Additional context

_No response_

triphora:
image
Not bothering to blur any of the sensitive data out because it's a local DB.

This was a PITA to figure out and test, but it seems to work as expected locally.

### DIFF
--- a/src/auth/validate.rs
+++ b/src/auth/validate.rs
@@ -7,6 +7,7 @@ use crate::models::users::{Role, User, UserId, UserPayoutData};
 use crate::queue::session::AuthQueue;
 use crate::routes::internal::session::get_session_metadata;
 use actix_web::HttpRequest;
+use base64::Engine;
 use chrono::Utc;
 use reqwest::header::{HeaderValue, AUTHORIZATION};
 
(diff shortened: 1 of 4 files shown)


src/auth/validate.rs
```diff
@@ -178,13 +180,33 @@ where
...
+            let credentials: String =
+                String::from_utf8(decoded).map_err(|_| AuthenticationError::InvalidCredentials)?;
+
+            Ok(credentials
+                .split_once(':')
+                .ok_or_else(|| AuthenticationError::InvalidCredentials)?
+                .1
+                .trim()
+                .to_string())
+        }
```
cyyynthia, line 207: It's kind of awkward and unexpected to have the client_id be in the request payload, and the secret extracted from the Basic authentication header. Having any of these be in the request body is not recommended as per the standard.

This would work for my generic implementation (since they're sent in both ways for compatibility reasons), but a client might expect to only send them through Basic authorization, or only through the HTTP payload.

I know it adds some complexity to the authentication routine, that's the blessing of standards that are 14 years old! 😇  My intuition would be to change the signature to output an enum, so a variant can hold 2 strings and pass them along to oauth checks.

PR #941, opened 2024-07-26
Fix 404 API spec link in README.md

FifthTundraG:
https://docs.modrinth.com/api-spec/ is no longer used for the API spec and returns a 404, but the link in the README still uses the old URL. This PR fixes that and switches it to https://docs.modrinth.com/.


FifthTundraG:
It's worth noting that https://docs.modrinth.com/api-spec _without_ a trailing backslash will redirect to the main page, but going to https://docs.modrinth.com/api-spec/ _with_ a trailing backslash will return a 404, because it's treated as a directory instead of a file. This is probably an oversight, but I'm uncertain whether or not this would be considered an issue with the documentation's redirects or would be considered a user error. Personally, I think that since a trailing backslash worked before the restructuring of the site, it's important to ensure compatibility with older documents that still may use this link alongside a trailing backslash, which would warrant fixing this issue. I figured I should ask if this is the right move before making a contribution, though 🙂

piprett (approved):

PR #949, opened 2024-08-12
Add caching statistics

Erb3:
I'll start by saying that the code here is probably not ideal, it is a lot of duplicate. I am unsure how to deduplicate it, as the SQL querries are different, and async. With that said, it appears to work well based on my testing. It caches the statistics for 30 minutes at a time, which will heavily improves the response times of this endpoint.

### DIFF
--- a/src/routes/v2/statistics.rs
+++ b/src/routes/v2/statistics.rs
@@ -1,7 +1,10 @@
-use crate::routes::{
-    v2_reroute,
-    v3::{self, statistics::V3Stats},
-    ApiError,
+use crate::{
+    database::redis::RedisPool,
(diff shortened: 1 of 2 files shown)


IMB11:
This would be great for #1301 on the monorepo.

Erb3:
> This would be great for #1301 on the monorepo.

Correct me if I am wrong, but modrinth/code#1301 seems to be focusing on analytics, not the global statistics?

IMB11:
Are the statistics not used by analytics endpoints?

IMB11:
Oh I misunderstood, this is just the general labrinth instance statistics.

